PR #4, opened 2015-08-03
Update file whitelist

cesarandreu:


AlexKVal:
I propose to migrate from `.npmignore` to `package.json` `files` section.

mtscout6:
I agree, managing a whitelist for releases is easier than managing a blacklist.

jquense:
sounds good to me. @cesarandreu you up for making that change?

cesarandreu:
Sure, I'll do that now

cesarandreu:
Just adding lib to files?

jquense:
I think that is all that's needed?

mtscout6:
We should also include the README and LICENSE otherwise they won't be on npm when we release. You can look here for an example: https://github.com/react-bootstrap/react-bootstrap/blob/v0.24.3/package.json#L27-L33

PR #5, opened 2015-08-03
[added] Better Modal state management

jquense:
Abstract out all the container styling and modal jumbling.

This lets us properly support having more then one modal open at at time, by removing all those assumptions.

We still need to think of a proper api to expose to Modal consumers, since they will need to know certain info like the modal's position in the overall stack for styling backdrops correctly. There is a longer term question of allowing a write-access api to the manager if folks want to implement more full blown windows, where they'd need to reorder the stack(s).

__bonus__ we could make this all pluggable so if folks really want to get into it they can just provide their own manager

### DIFF
--- a/examples/Modal.js
+++ b/examples/Modal.js
@@ -2,28 +2,37 @@ import React from 'react';
 import Button from 'react-bootstrap/lib/Button';
 import Modal from 'react-overlays/Modal';
 
-const ModalStyle = {
+let rand = ()=> (Math.floor(Math.random() * 20) - 10);
+
+const modalStyle = {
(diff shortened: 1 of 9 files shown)


jquense:
woohoo finally made it green

taion:
That's quite cool. What's the purpose of applying and removing the random top/left shifts in transform v left/top?

jquense:
ahh that's so the example modals don't sit exactly on top of each other, which is a poor demostration of stacking, since you couldn't see the one below.

taion:
Oh, that's the example. Hah. Misread, sorry.

PR #8, opened 2015-08-05
Remove extraneous dev dependencies

AlexKVal:
'karma-shinon-chai' succeeds both
'karma-chai' and 'karma-sinon'


jquense:
does this work? sinon chai was failing on me...i guess it does since the tests are green

AlexKVal:
At first I implemented it too in `R-B` https://github.com/react-bootstrap/react-bootstrap/pull/751,
and then I removed it https://github.com/react-bootstrap/react-bootstrap/pull/771 :smile: 

A quote from https://github.com/react-bootstrap/react-bootstrap/pull/771:
With babel we can do it like this now:
Instead of
`let props = Object.assign({}, this.props);`
use
`let props = {...this.props};`
which is transpiled into
`var props = _extends({}, this.props);`


What do you think ?

--
I with no hesitance will remove `object.assign()` part from this PR - if you still want it of course :cherries:

jquense:
Ya, most times the spread operator is good enough and built into babel. The reason I keep it around is for situations when I want to do `Object.assign(someObj, anotherObj)` where I need to mutate `someObj` instead of creating a new one, which you can't do with the spread syntax

AlexKVal:
Ah.. didn't know about this.
I removed the second commit about `object.assign`.

PR #11, opened 2015-08-11
[changed] focus target of the modal to its content

jquense:
Not sure about the best way to handle this.

We need to focus _something_ but focusing the wrapping div doesn't work great for bootstrap modals. This method does work, but it opens the possibility of a modal implementer not adding a tab index. The warning feels like a weak attempt to fix that,

Alternatively we could have a `focus` prop or something so the user can focus whatever they want, or just find the first focusable element and use that?

cc: @taion @mfunkie

### DIFF
--- a/package.json
+++ b/package.json
@@ -92,6 +92,7 @@
   "dependencies": {
     "classnames": "^2.1.3",
     "dom-helpers": "^2.2.4",
-    "react-prop-types": "^0.2.1"
+    "react-prop-types": "^0.2.1",
+    "warning": "^2.0.0"
   }
 }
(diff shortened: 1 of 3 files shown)


joemcbride:
If I'm remembering correctly, the focus needs to go on the div with the `aria-label` for accessibility, though the Modal is missing that label last I looked.

jquense:
@joemcbride some of that depends on what the roles are of each element as well. I am pulling out the screen reader to do some more thorough testing...

jquense:
So my research suggests that it doesn't matter what element we focus b/c we can't set `role='dialog` anyway, since it breaks SR navigation of the modal by placing it in another focus mode. In terms of labeling we, need to rely on the user to specify a label in the first place so we can jsut demostrate setting it on the content instead of the Modal component.

jquense:
gonna merge for the sake of continued iteration, but if anyone has more feedback about the api, or api alternatives, lets still talk it out

Issue #12, opened 2015-08-11
accessibility


jquense:
There are a few issues here, that are not trivial to fix, to make the modal actually useful to a SR and not just what upstream bootstrap shows as accessible.

This is really made complicated by both stacking, and multiple containers.

1. when a modal is open `aria-hidden` should be applied to the rest of the document content, so a SR can't shift virtual focus to it (which is separate from actual focus which is already handled). Essentially we need to create a virtual backdrop for the SR, so that it can't interact with anything else. This only actually makes sense tho when the container is the document body, or some other top level root. If the container is some deeper component, adding aria-hidden to it only hides some of the content.

2. On top of this, any additional modals _with backdrops_ that open should cause existing modals to be aria-hidden.

This gets sort of messy pretty quick. I'd like to handle it if we can, since doing this correctly is not super easy for a consumer, but I don't want to add a ton of overhead either.

jquense:
I think we can simplify this by just assuming that when we say "modal" we actually mean modal, ie. you can't interact with anything behind it

PR #13, opened 2015-08-11
Aria hidden

jquense:
related to #12 

This is a first pass at doing _some_ aria-hidden stuff. All of this is pretty ugly, from a component perspective. Unfortunately doing this is really necessary for a modal to appear like a modal to a screen reader, Without the aria-hidden on the rest of the page content the reader will shift focus (without regard for our focus trapping code) to the rest of the page.

This is rendered very useless though, if the `container` isn't a parent of the application, where modals would be its siblings.


src/utils/manageAriaHidden.js
```diff
@@ -0,0 +1,31 @@
...
+
+  [].forEach.call(container.children, node => {
+    if (mount.indexOf(node) === -1 && isHidable(node)){
+      cb(node);
+    }
+  });
+};
+
+export function ariaHidden(show, node){
+  if (!node) {
```
AlexKVal:
```js
if (!node || !isHidable(node)) {
```
?

taion:
That caveat regarding body is probably fine, right? If your container isn't the body, then even the visible backdrop doesn't get rendered over everything - just over the container.

jquense:
yes but that also assumes that the modal will be confined to the container with some css. the case I am sort of thinking of is having like a div off the body where everything is appended... We can't really solve all the issues I guess, we can at least suggest in the docs that folks just use the document body as the container.

taion:
(I don't understand why you'd do that. It seems like extra work for no good reason. I don't understand contained modals at all anyway, though.)

jquense:
why do we event allow that for modals? We should reexamine even showing people that in RB i think...

Issue #3, opened 2015-08-03
Allow react 0.14.0-beta1 as peer dependency


cesarandreu:
Does anything break with react 0.14.0-beta1?

taion:
Everything to do with context will break. We'd need to use `unstable_renderSubtreeIntoContainer` or whatever they're calling it this week.

cesarandreu:
Oh, I wasn't aware. I don't expect I'll be using contexts with my modals or popovers, so it should be alright.

In the meantime I made a branch [react-0.14](https://github.com/cesarandreu/react-overlays/tree/react-0.14) which replaces React with ReactDOM. It removes incompatible modules so I'm fairly certain that docs won't build... But since it's just for myself, it's alright for now. At least all the tests pass. I had to update a test, because findRenderedDOMComponentWithTag only accepts composite component.

I made another branch in which I added the built files: [react-0.14-build1](https://github.com/cesarandreu/react-overlays/tree/react-0.14-build1). In case someone wanted to play with it. 

No clue how people will be able to make libs that work with both v0.13 and v0.14...

jquense:
ya more then likely it won't be possible to maintain compat across them, which leaves us awkwardly waiting until everyone supports it.. I'm holding off a on major bumping a few of my libs just to not have to do it again when 14 comes out.

AlexKVal:
Resolved by #16

PR #15, opened 2015-08-17
Support react v0.14.0

cesarandreu:
DO NOT MERGE

Just sending this PR to increase visibility, issue #3


cesarandreu:
@taion you said this would break context, so it should use `unstable_renderSubtreeIntoContainer`. Is it simply a matter of replacing all instances of `ReactDOM.render` with `ReactDOM.unstable_renderSubtreeIntoContainer`?

jquense:
ya, you should just be able to replace Portal render calls with it

jquense:
and actually, maybe I can make a seperate branch to PR this against, that way we can release a prerelease version for 0.14 support like we did in react-bootstrap.

jquense:
https://github.com/react-bootstrap/react-overlays/tree/react-14

cesarandreu:
Gotcha. Added a new PR.

PR #17, opened 2015-08-20
DO NOT MERGE: Add OverlayTrigger

cesarandreu:
I needed it, so I ripped it out from react-bootstrap.
This needs to get cleaned up, but I figure having a PR doesn't hurt.

The examples server wasn't working for some reason, so I converted the two loaders to ES5 and used webpack-dev-server from the CLI.

Using npm v3.3.0, I managed to install all dependencies and generate a new example bundle. I was getting errors and updating all outdated dependencies somehow solved it. Re-added the jquense/component-playground package to get stuff working again.

In OverlayTrigger I removed the reference to react-bootstrap from the warning message. In OverlayTriggerSpec I removed the overlay types, since there's no tooltips / popovers here. Probably need to add a new test-case for that. 

Reduced the Transition timeout from 5 seconds to 1 second, because otherwise the test would time out. Increased the wait-time for that test to 5 seconds as well.

You can confirm that it works by checking out by branch, and running a file-server from the examples folder.


cesarandreu:
@jquense @taion What do I need to do in order to get this merged?

taion:
How did this change 16k lines???

Actually my first thought was what @jquense mentioned at https://github.com/react-bootstrap/react-bootstrap/pull/1174#issue-101433766 about overlapping components.

package.json
```diff
@@ -42,24 +42,25 @@
...
     "babel-plugin-object-assign": "^1.2.1",
     "chai": "^3.2.0",
-    "css-loader": "^0.15.6",
+    "component-playground": "github:jquense/component-playground",
+    "css-loader": "^0.16.0",
     "es5-shim": "^4.1.7",
-    "eslint": "^0.24.0",
-    "eslint-plugin-babel": "^1.0.0",
-    "eslint-plugin-mocha": "^0.4.0",
+    "eslint": "^1.2.0",
```
taion: This actually pretty much disables all the `eslint` rules, because of the changed defaults between 0.24 and 1.0.

src/OverlayTrigger.js
```diff
@@ -0,0 +1,265 @@
...
+  toggle() {
+    if (this.state.isOverlayShown) {
+      this.hide();
+    } else {
+      this.show();
+    }
+  },
+
+  componentDidMount(){
+    this._mountNode = document.createElement('div');
```
taion: Well, now that we're moving this into a new package, we should really document that this is creating a node that isn't inserted into the DOM, and that we do so to avoid... inserting extra nodes into the DOM. I found it confusing initially but forgot to ask to get a comment added here.

src/OverlayTrigger.js
```diff
@@ -0,0 +1,265 @@
...
+    this._overlay = this.getOverlay();
+
+    props.onClick = createChainedFunction(trigger.props.onClick, this.props.onClick);
+
+    if (isOneOf('click', this.props.trigger)) {
+      props.onClick = createChainedFunction(this.toggle, props.onClick);
+    }
+
+    if (isOneOf('hover', this.props.trigger)) {
+      warning(!(this.props.trigger === 'hover'),
```
taion: Wonder if we can make this a `propTypes` thing.

taion:
Going to let @jquense comment since I'm way behind on catching up to the most recent changes on this repo... just added a few unwarranted bikesheddy comments.

taion:
Does this need to also be backported to the non-React 14 branch?

dozoisch:
@taion it's because `examples/static/bundle.js` is commited

jquense:
I don't have a chance to do a real technical review but I do want to talk about whether this actually makes sense here. Would we remove it from RB itself? I sort of feel like this is too specialized a component for here, but i historically don't like this component to begin with :P 

Practically I am a bit concerned about the possible churn moving this here might introduce only to just re-export in RB, I don;t know that it makes sense to remove _all_ generic things from there, but i'm open to others thoughts about it.

src/OverlayTrigger.js
```diff
@@ -0,0 +1,265 @@
...
+        'Specifying only the `"hover"` trigger limits the visibilty of the overlay to just mouse users. ' +
+        'Consider also including the `"focus"` trigger so that touch and keyboard only users can see the overlay as well.');
+
+      props.onMouseOver = createChainedFunction(this.handleDelayedShow, this.props.onMouseOver);
+      props.onMouseOut = createChainedFunction(this.handleDelayedHide, this.props.onMouseOut);
+    }
+
+    if (isOneOf('focus', this.props.trigger)) {
+      props.onFocus = createChainedFunction(this.handleDelayedShow, this.props.onFocus);
+      props.onBlur = createChainedFunction(this.handleDelayedHide, this.props.onBlur);
```
taion: Not your bug, but see: https://github.com/react-bootstrap/react-bootstrap/issues/1140#issuecomment-133763740

Assuming we get this fixed in R-B first, we'll have to backport the (mostly trivial) fix.

taion:
I think overlay triggers are generally useful. It's pretty annoying to use tooltips without them, and I don't think we should force our users to invent their own.

Is this literally the exact same as the R-B `OverlayTrigger` though? I'd like for the marginal R-B user to be able to just use the package and not have to think about react-overlays at all, but I'd also like to not just do straight re-exports.

Difficulties.

jquense:
as far as I can see/think of, it would be exactly the same. to my mind the overlay trigger is the thing in RB that is doing most of the implementing for the twbs tooltip/popover logic. react let's us split them in an more decoupled way but ultimately I think that Component is really the js part of the it. Which is to say it may be better off in RB, if only to reduce the maintenance burden. I dunno. it's tough

cesarandreu:
Sorry for the delayed response.

@taion it's exactly the same, with a tweaked test.

I think it's really useful! I have a lot of informational tooltips, popovers, and such. I'd prefer having it in react-overlays, because otherwise I'm probably just gonna end up re-implementing it myself in my app. I'd assume others are on the same (or similar) boat.

Is there any pitfall to doing re-exports in RB?

taion:
We wanted to avoid doing so for the "utility" parts of this library like `Portal`, `Position`, `Overlay`, &c.

I think we have to re-export `OverlayTrigger` though.

jquense:
Here is what I think is my position. Tooltips and Popovers are a core component to Bootstrap. The main functionality of the plugin is to show a tip next to a trigger target. OverlayTrigger, while generically useful, contains the all of the public api of the bootstrap plugins, We definitely have to at least reexport it. However, since its a core component I feel weird have to have folks go to a different library to even view the code for it. Given that, and the the other fact that you can still just import _only_ the Overlay trigger from `react-bootstrap/lib` removing from RB does marginally more "harm" to our users than not having it in react overlays does.

I hold all this fairly loosely, @taion I leave it up to you for the final call though

jquense:
I just thought of a practical reason this is hard actually. `Overlay` in RB is different from the one here, and OverlayTrigger in RB depends on the RB Overlay not the react-overlays Overlay.

taion:
Even better!

I think I'm okay with just doing something like what we've done for `Overlay`, then. Users already can't really see how the overlays work without going to the react-overlays library, so it doesn't seem like a huge jump to do the same for `OverlayTrigger`.

PR #20, opened 2015-08-28
Allows nesting root close wrappers

jquense:
ran into a need with bootstrap nested dropdowns

### DIFF
--- a/src/RootCloseWrapper.js
+++ b/src/RootCloseWrapper.js
@@ -8,11 +8,19 @@ import ownerDocument from './utils/ownerDocument';
 // TODO: Consider using an ES6 symbol here, once we use babel-runtime.
 const CLICK_WAS_INSIDE = '__click_was_inside';
 
-function suppressRootClose(event) {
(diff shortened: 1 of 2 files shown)


jquense:
@taion does this seem reasonable?

src/RootCloseWrapper.js
```diff
@@ -21,6 +29,11 @@ export default class RootCloseWrapper extends React.Component {
 
     this.handleDocumentClick = this.handleDocumentClick.bind(this);
     this.handleDocumentKeyUp = this.handleDocumentKeyUp.bind(this);
+
+    let { id, suppressRootClose } = getSuppressRootClose();
+
+    this._suppressRootId = id;
+    this._suppressRootClosehHandler = suppressRootClose;
```
taion: typo :D

taion:
Yeah this is fine I think. The counters will keep incrementing when rendering on server, but I don't think `RootCloseWrapper`s ever get implemented on the server side, so no real issue.

With React 14 you could use context for this if you were so inclined, but... eh, why bother.

PR #22, opened 2015-09-01
Set default top and left position

medja:
This should help prevent overflow on some elements while calculating the target's position.
I found this to be useful when using Overlay if the new element forces the browser to display a scrollbar.
Without setting a position static elements shouldn't be affected, but this might cause problems for elements with a relative position.


taion:
Sorry it took so long to get to this. Thanks for the fix!

PR #29, opened 2015-10-03
Add support for setting the 'className' used on a Portal/Overlay component.

HPate-Riptide:
I ran into an issue where I needed to namespace some CSS when using react-bootstrap and I couldn't find an easy solution for the `Popover` component, where a `Portal` could add it to the DOM outside of the regular React root component.  

This pull request adds the ability to set a `className` on the `Portal` component when using a `Portal` or an `Overlay`.

```javascript
<Portal className="test-portal-class" container={()=> this.refs.container}>
    <span>Hello World</span>
</Portal>
```

or with an `Overlay`
```javascript
<Overlay
  show={this.state.show}
  container={this}
  target={ props => React.findDOMNode(this.refs.target)}
  portalClassName="test-portal-class"
>
  <ToolTip>
    Hello World
  </ToolTip>
</Overlay>
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,3 @@
-lib/
-
 # Logs
 logs
 *.log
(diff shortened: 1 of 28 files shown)


taion:
?

PR #30, opened 2015-10-28
[WIP] [added] Affix

taion:


src/Affix.js
```diff
@@ -0,0 +1,185 @@
...
+
+    this._isMounted = false;
+  }
+
+  onWindowScroll() {
+    this.onPositionUpdate();
+  }
+
+  onDocumentClick() {
+    requestAnimationFrame(() => this.onPositionUpdate());
```
jquense: recalculate on document click? is that just sort of a shotgun mutation observer :P or is there a more subtle thing I'm missing here?

src/utils/getDocumentHeight.js
```diff
@@ -0,0 +1,13 @@
...
+ *
+ * @returns {documentHeight: number}
+ */
+export default function (doc) {
+  return Math.max(
+    doc.documentElement.offsetHeight || 0,
+    doc.height || 0,
+    doc.body.scrollHeight || 0,
+    doc.body.offsetHeight || 0
+  );
```
jquense: LOL web tech

src/Affix.js
taion: Upstream! No clue why.

src/utils/getDocumentHeight.js
taion: `document.height` is `undefined` on Chrome so this always returned `NaN`. This is actual code in React-Bootstrap and I don't know why it isn't broken.

src/Affix.js
```diff
@@ -0,0 +1,185 @@
...
+  }
+
+  onPositionUpdate() {
+    if (!this._isMounted) {
+      return;
+    }
+
+    const {offsetTop} = this.props;
```
jquense: So tbs also allows functions for the offsetValues, does it make sense to try and support that here as well?

src/Affix.js
```diff
@@ -0,0 +1,185 @@
...
+
+    if (positionTopMin > this.getPositionTopMax()) {
+      this.setState({
+        affixed: 'bottom',
+        affixStyle: {
+          position: 'absolute'
+        }
+      }, () => {
+        this.setState(({affixed}) => {
+          if (affixed === 'bottom') {
```
jquense: might need another `_isMounted` check here...

src/Affix.js
taion: I don't think it's needed - since it's React the parent can just recalculate.

jquense:
looks good I'm still traveling but if you want to merge go for it. I'm gonna do some work with affix for work next week so I can put it through its paces then

taion:
A few more things to do here -

1. Optimize re-rendering
2. Add tests (ugh)
3. Add an example (how do you demo an affix??!)

And then replace R-B's busted Affix with this for completeness (though the one there will be almost entirely pointless).

test/AffixSpec.js
```diff
@@ -0,0 +1,254 @@
...
+  // This makes the window very tall; hopefully enough to exhibit the affix
+  // behavior. If this is insufficient, we should modify the Karma config to
+  // fix the browser window size.
+  class Container extends React.Component {
+    render() {
+      return (
+        <div style={{
+          position: 'absolute',
+          top: 0,
+          height: 20000
```
taion: Hard-coding large numbers FTW

taion:
Okay, I have examples and tests now. Do you want to take a look to make sure I'm not doing something totally dumb before I merge? I'm going to be running off my local version of react-overlays anyway for the time being, so I'm not blocked on getting this merged.

jquense:
I'm for merging maybe we hold off a bit on cutting the react-bootstrap one until we've had a chance to test it in some real situations a bit? Positioning stuff is always a lot uglier in real life.

Which is to say I don't mind bumping versions often or quickly in this repo

taion:
It's _almost_ ready, but the auto `offsetTop` is really glitchy right now. I'm going to try to figure out if there's a better way to handle it.

taion:
I can't really tell whether https://github.com/react-bootstrap/react-overlays/commit/7c0771a2b54556ffaa691d09279aad7b2b17c455 makes things better or worse. Maybe I should give up on automatically handling `offsetTop`.

taion:
The strategy here is to add a dummy element that I can use to figure out where the Affix should go. This solves part of the problem, but now I need to figure out a way to listen for this dummy element moving around in the document (e.g. if it's below a header that's changing size or something). Do you know if there's a good way to do this?

taion:
Maybe a better approach is to just add an `AutoAffix` component at some other point - I don't really enjoy having to pass in `offsetBottom` either.

If I do all of these things, plus listen to resize events, `AutoAffix` should cover most (but not all) use cases.

jquense:
It seems reasonable to split it into two components yeah.

Where is the trouble coming from? affixing or inlining the affixed element? If its inlining the element again, maybe just cache the initial offset when it was affixed and just use that? IT doesn't cover if the offset has changed but it may be simpler just t include that caveat. otherwise seems like you need a dummy element and maybe a mutation observer?

taion:
@jquense 

The issue is exactly keeping track of where the scroll offset for the transition should be, in cases where it changes, i.e. keeping the element inlined when it should be.

My first cut has the cached offset, but this breaks even in really simple cases like just resizing the window width on a responsive page.

Although looking at the affix on the TWBS docs site, the one they have glitches out as well after window resizing, so maybe I don't need to try quite so hard.

taion:
Success! `<Affix>` is a reasonable base component, while `<AutoAffix>` gives behavior pretty close to upstream TWBS for cases where you don't want to do math.

taion:
This took so much longer than I wanted, but I think `AutoAffix` is actually just good enough for most cases - or at least as good as e.g. what's demonstrated on the TWBS docs pages.

PR #31, opened 2015-11-01
Fix transition in behaviour for unmountOnExit

martynsmith:
It seems that if you're using the `unmountOnExit` option, then the Transition component both mounts your child component _and_ applies the `enteringClassName` in the same Javascript tick, this causes the browser to not trigger the animation (and instead the `enteringClassName` attributes are immediately applied).

By forcing a wait until the next Javascript tick in this case animation appears to reliable happen.

### DIFF
--- a/src/Transition.js
+++ b/src/Transition.js
@@ -72,7 +72,7 @@ class Transition extends React.Component {
       // EXITED is always a transitional state to either ENTERING or UNMOUNTED
       // when using unmountOnExit.
       if (this.props.in) {
-        this.performEnter(this.props);
+        setTimeout(() => this.performEnter(this.props), 0);
       } else {
         this.setState({status: UNMOUNTED});
       }


jquense:
Hi there, thanks for the PR,

I am a bit reticent to take this, since generally its the animation implementer's job to [reflow the DOM](https://github.com/react-bootstrap/react-bootstrap/blob/master/src/Collapse.js#L97) at the points that are appropriate for the given animation. Otherwise we may be introducing an unnecessary delay (albeit a tiny one). Is there a reson why a similar strategy wouldn't work for your use case, using the `onEnter`hook?

martynsmith:
Yes, I'm definitely not an expert in these matters, but I do think it should be part of the library to ensure this works (since it's providing the `unmountOnExit` option, it seems only reasonable that it would work correctly).

I'm certainly not wedded to my patch, but it would be nice if there were _some_ mechanism for the library to ensure the browser had rendered the newly added element before applying the `enteringClassName` class.

Do you want me to try reworking the patch to use the reflow thing you linked to?

jquense:
yeah I am not sure myself... @taion any thoughts?

taion:
I'm not sure I understand what's going on. The lifecycle of going from unmounted to entered is supposed to be:

1. Set state from `UNMOUNTED` to `EXITED` in `componentWillReceiveProps`: https://github.com/react-bootstrap/react-overlays/blob/v0.5.0/src/Transition.js#L51-L55
2. Component renders with `EXITED` state: https://github.com/react-bootstrap/react-overlays/blob/v0.5.0/src/Transition.js#L176-L177
3. Start transitioning the component in from `componentDidUpdate`: https://github.com/react-bootstrap/react-overlays/blob/v0.5.0/src/Transition.js#L71-L75
4. Subsequent re-renders.

The React docs say that `componentDidUpdate` only gets called after the results of `render` are flushed to the DOM: https://facebook.github.io/react/docs/component-specs.html#updating-componentdidupdate

Is there something weird happening here?

jquense:
My assumption is that while the change is flushed, it doesn't mean that all renders occur on different ticks, and depending on what is actually flushed to the DOM it may not trigger a reflow/repaint. TBH I have a hard time understanding how that could be, but this part of css animation has always been fraught so it wouldn't surprise me...

As a point of potential evidence, the Collapse transition has to make sure it reflows the DOM at specific points otherwise it'll just not animate, this is in spite of React running though the render cycle a few times

taion:
It sounds like the right thing to do then is to wrap both the `componentDidMount` and `componentDidUpdate` handlers with `requestAnimationFrame`, then, and tie it into the cancellable-next-callback thing.

jquense:
RAF is going to be too slow here I think (ok probably not actually noticeably), we don't need classes to be added when the browser refreshes, the css animation should already handle that correctly.

If we _need_ to do something it should be enough to just trigger a `get` on some node property like `offsetHeight`, like we do in Collapse. Ultimately the problem isn't actually that they occur in the same tick, its that the Browser flushes the change at the same time, so the animation is skipped. But is that better left to the user in `onEnter`?

jquense:
What do we think about this @taion?

taion:
I don't think `setTimeout` is right here.

I guess the question is – does it make sense to always trigger a reflow here, or should each individual animation do it per your comment https://github.com/react-bootstrap/react-overlays/pull/31#issuecomment-152870135?

jquense:
In general I think doing potentially unncessary reflows during an animation is not a good default, as a single component it makes sense to do this by default on mount, but if you have a bunch of this animating on screen I am not at all sure what whether that's still a good default.

taion:
True – but how many things do you expect people to have that are transitioning in on mount at the same time?

If you have enough components that are configured this way, your page is going to be obnoxious as all hell anyway.

martynsmith:
The general idea behind it was for things like bootstrap modals, I don't render them the whole time, just mount them when I want to display them (but at the moment the animation doesn't work).

PedroFonseca:
I've been trying to figure out how to render the show transition properly for a while now and i got to the same solution as Marty... I get that you don't want the delay when the transition isn't needed, so why not just provide an additional property for when we need this delay? Or maybe even do it automatically when the unmountOnExit property is true.

taion:
This was fixed in https://github.com/react-bootstrap/react-overlays/pull/60 I think.

PR #34, opened 2015-11-15
Forward className and style to Affix child

taion:


taion:
On second thought, never mind. Should just style the child directly.

PR #36, opened 2015-11-16
[fixed] AutoAffix nnot passing width or updating

jquense:


src/AutoAffix.js
```diff
@@ -119,12 +133,12 @@ class AutoAffix extends React.Component {
         <div ref="positioner" />
 
         <Affix
+          {...props}
```
taion: Oops.

I think I wanted to let the user override `offsetTop` or something via `props`, but on second thought that's a silly thing to do.

src/AutoAffix.js
```diff
@@ -32,6 +32,11 @@ class AutoAffix extends React.Component {
     this._windowScrollListener = addEventListener(
       ownerWindow(this), 'scroll', () => this.onWindowScroll()
     );
+
+    this._windowResizeListener = addEventListener(
```
taion: Should probably add something like this to `<Affix>` as well. I thought about doing this originally - it's one of those things that's the right thing like 90% of the time, but then you end up needing something much heavier like https://github.com/souporserious/react-measure to handle the rest of the cases. But this might be the best option - the `MutationObserver` polyfill for IE9/10 sorta sucks.

taion:
:+1:

src/AutoAffix.js
jquense: yeah I figured since we aren't getting every case it couldn't hurt to catch a few more cases. makes more sense with autoWidth

PR #37, opened 2015-11-17
[added] affix state callbacks

jquense:
So this doesn't actually fire anything for affixTop is it doesn't change after the first render, not sure if that is ok...

### DIFF
--- a/src/Affix.js
+++ b/src/Affix.js
@@ -127,7 +127,18 @@ class Affix extends React.Component {
       return;
     }
 
-    this.setState({affixed, position, top});
+    let upperName = affixed === 'affix'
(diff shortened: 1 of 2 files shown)


src/Affix.js
```diff
@@ -27,6 +27,7 @@ class Affix extends React.Component {
     };
 
     this._needPositionUpdate = false;
+    this._needPositionUpdate = false;
```
taion: ?

taion:
:+1: other than the duplicated line

src/Affix.js
jquense: weird typo

PR #39, opened 2015-11-19
[fixed] Modal does not fire show callback

nickuraltsev:


jquense:
Thanks for the PR, its appreciated.

Slight issue, the call to the callback should be in `componentDidMount` to be consistent with how it's fired the rest of the time in `componentDidUpdate` also I am fairly sure that this will cause the callback to be fired twice on normal updates, one in `onShow` and again in `componentDidUpdate`

nickuraltsev:
The callback is called in `Modal#onShow` which, in its turn, is called in `componentDidMount` and `componentDidUpdate`. ModalSpec.js contains two tests that verify that the callback is called exactly once on mount and exactly once on each update:
* Should fire show callback on mount
* Should fire show callback on update

sompylasar:
@jquense @nickuraltsev What blocks this PR from merging? Thanks.

jquense:
The holidays mostly, and that I forgot about this.

sompylasar:
OK, thanks.

jquense:
thansk for the PR!

nickuraltsev:
You are welcome!

PR #42, opened 2015-11-21
Add dynamicPosition prop to Position component

thetimbanks:
This adds the `dynamicPosition` prop to the `Position` component in order to allow for the popover to be repositioned when content changes. It is turned off by default so it should not affect any existing code.

This goes with this issue on react-bootstrap: https://github.com/react-bootstrap/react-bootstrap/issues/1438

### DIFF
--- a/src/Overlay.js
+++ b/src/Overlay.js
@@ -30,6 +30,7 @@ class Overlay extends React.Component {
       , containerPadding
       , target
       , placement
+      , shouldUpdatePosition
       , rootClose
       , children
       , transition: Transition
@@ -48,7 +49,7 @@ class Overlay extends React.Component {
     // Position is be inner-most because it adds inline styles into the child,
     // which the other wrappers don't forward correctly.
     child = (
-      <Position {...{container, containerPadding, target, placement}}>
+      <Position {...{container, containerPadding, target, placement, shouldUpdatePosition}}>
         {child}
       </Position>
     );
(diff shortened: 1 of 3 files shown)


thetimbanks:
Not sure why the tests are failing on TravisCI. When I run the karma tests locally everything passes.

taion:
That was a Travis misconfig on our part. This looks good, except the name for the prop is a bit clunky. Not a blocker but does anybody have any better ideas?

thetimbanks:
Good to know I didn't break anything!

I had some trouble naming the prop. Another idea I had was autoPosition.
Not much better but something different.

psachs21:
I'd like to see this change merged so maybe a name like forceRepositionOnRender would be more accurate? It's a little verbose but better verbose than cryptic.

thetimbanks:
Any other ideas on the name change for this property?

taion:
Could just call it `shouldUpdate`.

taion:
Well, then we'd have to manually rename it to `shouldUpdatePosition` on other components. Maybe `shouldUpdatePosition` then? It's consistent w/e.g. the naming on `<StaticContainer>`.

thetimbanks:
Prop has been renamed to `shouldUpdatePosition` and commits squashed.

taion:
LGTM. @jquense?

jquense:
LGTM

hellpf:
When it will be released?

jquense:
released

PR #43, opened 2015-11-24
[WIP]Allow container prop of Position component to be not only mountable but also function

haradakunihiko:
Currently, container property of Position component should be mountable, but it can be function, like Portal.js bellow.

https://github.com/react-bootstrap/react-overlays/blob/master/src/Portal.js#L21

### DIFF
--- a/src/Position.js
+++ b/src/Position.js
@@ -135,7 +135,10 @@ Position.propTypes = {
   /**
    * "offsetParent" of the component
    */
-  container: mountable,
+  container: React.PropTypes.oneOfType([
+    mountable,
+    React.PropTypes.func
+  ]),
   /**
    * Minimum spacing in pixels between container border and component border
    */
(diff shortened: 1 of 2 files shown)


taion:
Can you add a test case here?

haradakunihiko:
@taion Thanks, I was trying to look for the reason why the test failed. I will add the test case later.

taion:
Just some weird linting stuff, it wasn't a problem on your end.

taion:
@haradakunihiko Any update?

haradakunihiko:
@taion Sorry for the late reply. 
Added test case to check callback function works with container prop. It tests only for callback function to work and does not test about its use case such as using refs for container or something like that. If it's necessary, please tell me.

I'm checking the Travis build failure because it does not fail in my local environment.

taion:
Try rebasing your commits.

PR #46, opened 2015-11-27
Update linting dependencies to fix linting

taion:
This fixes all the broken builds. Hopefully `eslint:recommended` is good enough? If not I can try just upgrading babel-eslint.


AlexKVal:
"eslint:recommended" seems a good enought minimum.
If there are any concerns I could implement additional rules from the "old behavior" list
on this page http://eslint.org/docs/user-guide/migrating-to-1.0.0.html

taion:
@jquense You okay with this?

jquense:
yup! lgtm

PR #48, opened 2015-11-29
Update Travis config

taion:


taion:
This disables push builds, which just end up being bad dupes of PR builds anyway.

PR #49, opened 2015-12-02
Can use (but probably shouldn't) multiple modals at once

### LINKED ISSUE #41: `Uncaught TypeError: node.removeAttribute is not a function` with multiple modals
Example (adapted from official react-overlay modal example, with two modals at same level): 
https://gist.github.com/jesenko/e453679c1f1579bd5127

On close, error is thrown at `node.removeAttribute('aria-hidden');` in `react-overlays/lib/utils/manageAriaHidden.js`

donatasbacius:
Fixes #41. Please pull n publish this as atm react-bootstrap's modals, when used multiple at a time, are broken.

### DIFF
--- a/src/Modal.js
+++ b/src/Modal.js
@@ -161,7 +161,11 @@ const Modal = React.createClass({
     let show = !!props.show;
     let dialog = React.Children.only(this.props.children);
 
-    let setMountNode = ref => this.mountNode = (!ref || ref.getMountNode());
+    let setMountNode = ref => {
+      if (ref !== null && typeof ref === 'object') {
+        this.mountNode = ref.getMountNode();
+      }
+    }
 
     const mountModal = show || (Transition && !this.state.exited);
 


taion:
Can you please add a regression test and explain what exactly this code is supposed to do? In that context `ref` is the instance of the `<Portal>`, so I'm not sure how it could be `null`.

craigbilner:
@taion 

The null is explained by this:

https://facebook.github.io/react/docs/more-about-refs.html

>Note that when the referenced component is unmounted and whenever the ref changes, the old ref will be called with null as an argument

In terms of the logic, ```!ref || ref.getMountNode()``` shouldn't be used for mountNode as it will set a "node" to true for the null call (just for code clarity).

However, I think the better change would be to pass ```this.setMountNode``` as the "ref" which means the reference check made by React wouldn't think the reference had changed, which as far as I can tell hasn't and is only used to set the "mountNode"? By jumping out of the check then ref wouldn't be null (and would remove that less than ideal _this reference).

craigbilner:
@taion @donatasbacius 

written up here https://github.com/craigbilner/react-modal-play

jquense:
Sorry for the Delay folks, as @craigbilner is pointing out the proposed PR is actually not correct, `this.mountnode` _should_ be set to null when null. The problem is jsut my quick little psuedo conditional is logically incorrect.

@donatasbacius If you want to update the Pr, (or someone else submit one) I'll make sure it gets in and released quickly

src/Modal.js
```diff
@@ -161,7 +161,11 @@ const Modal = React.createClass({
     let show = !!props.show;
     let dialog = React.Children.only(this.props.children);
 
-    let setMountNode = ref => this.mountNode = (!ref || ref.getMountNode());
+    let setMountNode = ref => {
+      if (ref !== null && typeof ref === 'object') {
```
jquense: this really just needs to be: `this.mountNode = (ref ? ref.getMountNode() : ref)` otherwise the mountPoint won't actually get nulled out correctly

taion: Could just be `ref && ref.getMountNode()` no?

jquense: also that

craigbilner: And also move it into the class to avoid the ref check which was causing a similar but separate issue :-)

jquense: Sorry @craigbilner what do you mean "move it into the class"? What problem does having a new function cause on each render?

craigbilner: @jquense if you look at my investigations [here](https://github.com/craigbilner/react-modal-play) having a new ref (which isn't required) results in extra unnecessary work and was causing other side effects (which I didn't have time to re-create perfectly in a simple example). I couldn't see any benefit in spinning up a new function on every render?

craigbilner: also, should probably highlight this PR is a bit confusing as multiple modals are not a problem per se, I fixed this in our production code by weeding out unnecessary renders with `shouldComponentUpdate`.

jquense:
I've fixed this myself so closing this, thanks everyone for all the help :)

PR #51, opened 2015-12-05
Update tooltip components in examples

taion:


taion:
https://github.com/react-bootstrap/react-overlays/pull/45 was too zealous - `Position` actually should still forward `placement`.

Also, some non-React 14 compatible code in the examples.

PR #53, opened 2015-12-31
createChainedFunction supports for falsy

wuct:
The original `createChainedFunction` does not handle `undefined` properly. It throws when receiving `undefined`. This pull request will make `createChainedFunction` handle all falsy values properly.

### DIFF
--- a/src/utils/createChainedFunction.js
+++ b/src/utils/createChainedFunction.js
@@ -9,10 +9,10 @@
  */
 function createChainedFunction(...funcs) {
   return funcs
-    .filter(f => f != null)
+    .filter(f => f)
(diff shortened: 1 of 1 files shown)


taion:
`undefined == null`

You probably actually passed in `false`.

wuct:
Thank you @taion, your replying is so fast! I didn't know that `undefined == null`. 
However, don't you think this function will be more useful if it can handle `false`? In my case, I need to chained some functions conditionally. For example:

```
createChainedFunction(
  foo,
  shouldDoBar && bar
)
```

taion:
No. Do `shouldDoBar ? bar : null`.

PR #60, opened 2016-01-28
[fixed] New props are flushed before any new transition

### LINKED ISSUE #40: Transition can fail if new props affect animation state
Ran into this writing some Tabs stuff. Essentially as you switch tabs you apply an `'active'` class that sets the display to `block` this needs to happen _before_ the `in` class is added to the element to start the animation.

As it stands transitions triggered from a prop change actually start happening _before_ the new props are flushed to the DOM. ie triggered here: https://github.com/react-bootstrap/react-overlays/blob/master/src/Transition.js#L50-L63

my thought is that we need to move these calls to componentDidUpdate. 

cc @taion

jquense:
fixes #40.

### DIFF
--- a/src/Transition.js
+++ b/src/Transition.js
@@ -44,37 +44,49 @@ class Transition extends React.Component {
   }
 
   componentWillReceiveProps(nextProps) {
-    const status = this.state.status;
-    if (nextProps.in) {
-      if (status === EXITING) {
-        this.performEnter(nextProps);
-      } else if (this.props.unmountOnExit) {
-        if (status === UNMOUNTED) {
-          // Start enter transition in componentDidUpdate.
-          this.setState({status: EXITED});
-        }
-      } else if (status === EXITED) {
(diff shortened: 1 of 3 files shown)


src/Transition.js
```diff
@@ -44,31 +44,21 @@ class Transition extends React.Component {
...
+    else {
+      this._needsUpdate = true
     }
   }
 
   componentDidUpdate() {
-    if (this.props.unmountOnExit && this.state.status === EXITED) {
+    const status = this.state.status;
+
+    if (this.props.unmountOnExit && status === EXITED) {
```
taion: Do we need to `return` at the end of this and skip the bottom checks?

jquense: probably safer yeah...I'm honestly not sure though

taion: At a glance, it looks like you could at least hit `performEnter` twice, maybe.

PR #66, opened 2016-03-09
Portal container should be updated when container prop changes

tgriesser:
Test included, fixes the error `The node to be removed is not a child of this node.` occurring on unmount when the `container` prop was being updated but the `_overlayTarget` was not being moved in the document.

### DIFF
--- a/src/Portal.js
+++ b/src/Portal.js
@@ -32,6 +32,13 @@ let Portal = React.createClass({
     this._renderOverlay();
   },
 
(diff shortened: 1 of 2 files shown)


taion:
LGTM

spudly:
Any chance we could get a new release generated for this?

taion:
Done.

spudly:
Sweet, thanks!

PR #72, opened 2016-03-28
Add /src to files in package.json

tgriesser:
Makes it possible for a user to write a postinstall script which builds the /src directory when installing from a github release.

I was looking to get the latest patches working with react-bootstrap, but noticed it hasn't been published yet and it's not possible to npm install the latest github version and manually but that too doesn't work because /src isn't included.

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,6 +8,7 @@
   "main": "lib/index.js",
   "files": [
     "lib/",
+    "src/",
     "LICENSE",
     "README.md",
     "CHANGELOG.md"


taion:
Can you elaborate on what you mean? Don't you still need a fork anyway to have that postinstall script?

Also @jquense can you publish a new version? Or add me as a publisher and I'll do it?

jquense:
added to the npm package!

taion:
@tgriesser I'm going to cut a new release shortly. Are you okay if I close this? I think I'd rather we stay on top of cutting releases if we have critical fixes to deploy.

tgriesser:
Sure

taion:
Thanks. Just drop me a line if we end up falling behind a bit on releases. I'd rather we keep the released version up-to-date than have people stuck on non-released versions.

PR #74, opened 2016-04-03
Fix unmounting portal when parent is unmounted

tgriesser:
When a portal is mounted and references a container which unmounts before the portal, an error occurs as the result of `getContainerDOMNode` will become "body" and removeChild fails. This removes `getContainerDOMNode` in favor of caching the container on the instance as `_portalContainerNode`.

Test included.


taion:
Thanks!

taion:
Released as v0.6.2.

PR #82, opened 2016-04-28
add additional check to Modal componentWillUpdate, fixes #81

### LINKED ISSUE #81: Modal autoFocus does not refocus on last focused element when Modal is closed
I opened an issue previously here: react-bootstrap/react-bootstrap#1851

In short, the element triggering the modal is not being refocused when the modal is closed. It appears that what is happening is that `componentWillUpdate` is called twice, and in turn, `checkForFocus` is called twice. So the modal is opened, `checkForFocus` is called, `this.lastFocus` is set to the triggering element (good), `checkForFocus` is called again, and `this.lastFocus` is set to the modal (assuming that `autoFocus` is set to true). I don't know why `componentWillUpdate` is being called twice, but I double checked that my own component is being rendered only once. This issue also affects the modals in the react-bootstrap doc., so I don't think my own code is causing the issue.

A simple fix appears to be to check that both `nextProps.show` is true (which is the current check) and that `this.props.show` is false within `componentWillUpdate`. This prevents the call to `checkForFocus` occurring when it doesn't need to.

I'll submit a pull request in a few, in the mean time I'm curious to know if I'm just way off track here :)

vga-:


jquense:
LGTM

PR #90, opened 2016-05-19
Use getOffset regardless of element type

### LINKED ISSUE #89: Dom-helper container width should consider scrollbar
Hi, it seems like dom-helper is using `window.innerWidth` to get container width. This causes an incorrect offset computation in `Position` when the scrollbar is active. Are there any workarounds for it?

khankuan:
Fixes #89

### DIFF
--- a/src/utils/overlayPositionUtils.js
+++ b/src/utils/overlayPositionUtils.js
@@ -8,15 +8,12 @@ const utils = {
   getContainerDimensions(containerNode) {
     let width, height, scroll;
 
+    ({ width, height } = getOffset(containerNode));
     if (containerNode.tagName === 'BODY') {
-      width = window.innerWidth;
-      height = window.innerHeight;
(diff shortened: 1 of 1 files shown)


khankuan:
Will this bug be fixed?

jquense:
I don't think its a bug. Body is odd, as is document and window, that's why they are handled specially. bootstrap does the same thing: https://github.com/twbs/bootstrap/blob/v3.3.6/js/tooltip.js#L367

khankuan:
The problem is the separate logic for `BODY` does not consider scrollbar appearing or not :/

jquense:
I'm having a hard time understanding why that would cause an issue, can you put together a fiddle demonstrating the issue?

khankuan:
Nvm, it seems to be fixed in some way as i went back to debug. Thanks.

PR #93, opened 2016-06-06
[added] disabled prop to rootClose

jquense:
Allow the child to remain mounted while conditionally disabling the behavior

### DIFF
--- a/src/RootCloseWrapper.js
+++ b/src/RootCloseWrapper.js
@@ -61,15 +61,16 @@ export default class RootCloseWrapper extends React.Component {
       return;
     }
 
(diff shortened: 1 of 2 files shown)


taion:
I meant to ask – do you remember why we (I) did this instead of checking something like `domHelpers.contains(ReactDOM.findDOMNode(this), event.target)` or something?

jquense:
I don't remember hmm...this approach should be more performant, since its not doing a contains, tho contains is super fast actually. I don't know :P

taion:
It'd also skip the bizarre wrapper.

jquense:
probably worth revisiting

Issue #64, opened 2016-02-24
rootClose using React-Tap-Event-Plugin


Stenerson:
When using [react-tap-event-plugin](https://github.com/zilverline/react-tap-event-plugin) and `rootClose={true}` the overlay opens and closes immediately. I've spent most of the day in my Chrome debugger but I can't figure out why it works as expected when my target uses `onClick` and it closes immediately when my target uses `onTouchTap`.

Here's what my component's `render()` function is returning:

``` javascript
<button
  className={classnames("btn",{"btn-primary": this.state.open})}
  onTouchTap={this._toggleOpen}
  ref="target"
>
  Open Overlay
  <Overlay
    show={this.state.open}
    onHide={() => this.setState({ open: false })}
    rootClose={true}
    placement={this.props.placement}
    target={() => ReactDOM.findDOMNode(this.refs.target)}
  >
    /* Some Overlay content... */
  </Overlay>
</button>
```

Honestly I'm not sure if this is an issue with react-overlays or not. I did see an [old issue](https://github.com/react-bootstrap/react-bootstrap/issues/831) in react-bootstrap that indicated the React-Tap-Event-Plugin _should_ work so I thought I'd start here.

As a workaround I'm just using `onClick` for my Overlay triggers but as you know it has a very sluggish feel on mobile devices, especially since the rest of the app is using `onTouchTap`.

Any ideas? Thanks!

taion:
This sounds like an inconsistency between the component event handling and the backdrop event handling. Compare https://github.com/react-bootstrap/react-overlays/blob/92171fb806719aa37f063b46f174508fa254836f/src/RootCloseWrapper.js#L104 and https://github.com/react-bootstrap/react-overlays/blob/92171fb806719aa37f063b46f174508fa254836f/src/RootCloseWrapper.js#L51. Perhaps it ought to be possible to control the events injected in `<RootCloseWrapper>`?

dahjelle:
I believe I'm running into this issue while using [react-fastclick](https://github.com/JakeSidSmith/react-fastclick): the react-bootstrap `DropdownButton` opens and closes immediately. I'm able to reproduce this in Chrome's mobile device simulator where touch events rather than mouse events can be triggered.

In my case, the `react-fastclick` code rewrites `touch` events as `click` events when there is an `onClick` handler present. That part seems to work fine.

[...]

Issue #92, opened 2016-06-01
onmousedown vs onclick?


glenjamin:
Hello!

My main application is performing some actions on mousedown (like dragging).

The background overlay here only triggers onClick, this results in an effect where the overlays don't disappear when the user begins to interact with the application.

Would you be open to a PR which moves the background handlers onto `onMousedown` ?

taion:
ref https://github.com/react-bootstrap/react-bootstrap/pull/1786

glenjamin:
Seems related. Do you have any additional info about what problems that caused? Was the problem that the change should have been made over here, or something deeper?

jquense:
@glenjamin check out https://github.com/twbs/bootstrap/issues/13816, I'm not sure what issues where caused. If you want to do some research and see if they are no longer issues, we can make the switch

glenjamin:
Did a little bit more digging, the `Modal` doesn't really affect me at all - the overlay captures clicks and everything works as expected.

It's actually the `RootCloseWrapper` that I'm having trouble with, this binds to `click` on `document` to trigger the close - but allows the original event through as expected.

I did a quick trial of changing this to bind to `mousedown` and it worked as expected for me - but I'm unsure what else to check against.

jquense:
done?

PR #95, opened 2016-06-08
Make <RootCloseWrapper /> hide onmousedown instead of onclick

glenjamin:
As discussed in #92 

Unlike modal there's no background overlay, so click() is a bit late if the user is doing an interaction that starts onmousedown like dragging or clicking another button.

In order to run the examples from a fresh working tree I had to peg the component-playground version.

I've done some testing of this, I can't find any problems so far. The existing discussions about `mousedown` problems I've found have all been related to Modals and the background overlay.

### DIFF
--- a/src/RootCloseWrapper.js
+++ b/src/RootCloseWrapper.js
@@ -35,7 +35,7 @@ export default class RootCloseWrapper extends React.Component {
   constructor(props) {
     super(props);
 
-    this.handleDocumentMousedown = this.handleDocumentMousedown.bind(this);
+    this.handleDocumentMouse = this.handleDocumentMouse.bind(this);
     this.handleDocumentKeyUp = this.handleDocumentKeyUp.bind(this);
 
     let { id, suppressRootClose } = getSuppressRootClose();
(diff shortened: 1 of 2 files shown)


taion:
Thanks for the PR. To get this merged, let's make this configurable instead.

By default, most actions of this style do trigger on click – see e.g. the dropdowns on the GitHub page, or default TWBS dropdowns. As such, it's not correct to make our root close wrapper only trigger on mouse down.

glenjamin:
Sounds reasonable, what should the default be?

Some more data points: Gmail is mousedown, Google plus is click, Facebook is click, slack is mousedown

At a guess I'd say websites are more likely to be click, applications more likely to be down - although I think using mousedown over click doesn't really provide any/many downsides, but using click over mousedown can.

Any component that uses `<RootCloseWrapper />` would also need to be configurable then.

glenjamin:
There's a chance this might fix #64 as well

glenjamin:
The diff will be much more readable without whitespace. You can add `?w=1` on the querystring if you didn't already know.

jquense:
> You can add ?w=1 on the querystring if you didn't already know.

omg magic 🌟

taion:
BTW, where we got click in the first place was that this was what Bootstrap (v3) did (thought v4 does click as well).

@jquense pointed out that GitHub splits the difference – overlays close on click, but there's a backdrop-style element that prevents interaction with the rest of the page while the dropdown is shown.

So given all that I think the right default here is `click`.

glenjamin:
Reckon we're good to merge if I flip the default to click then?

We'd then need to add the prop to consuming components in React Bootstrap

taion:
I wonder if we can just hijack the existing `rootClose` prop to accept a string in addition to a boolean.

glenjamin:
This should be good to merge now, rebased off master and made `click` the default.

taion:
LGTM

PR #102, opened 2016-07-12
Make `target` work like `container`

taion:


taion:
Arguably this is breaking but anybody actually reading the `props` in the arguments to the target function is doing something utterly bizarre.

jquense:
i might the one person doing something bizarre, let me confirm there isn't a reasonable use case...

taion:
It seems really weird for a callback to be able to read the props of a different component.

jquense:
LGTM

PR #105, opened 2016-07-29
Upgrade react-prop-types

taion:


src/utils/requiredIf.js
```diff
@@ -1,11 +0,0 @@
-export default function requiredIf(propType, matcher){
```
taion: This was never used

PR #109, opened 2016-08-13
Enhance: Bind/Unbind root close handlers for RootCloseWrapper

Zapix:
Hi all. 

I've come here from https://github.com/react-bootstrap/react-bootstrap/issues/1808 
On those issue @taion asked to rewrite `RootCloseWrapper` which helps not to remount child components on `DropdownMenu`. The solution of react-bootstrap issue is properly binding/unbinding root close handlers. 

The idea was tested with react-bootstrap and it works! You could check it here (https://github.com/Zapix/react-bootstrap/tree/overlay-test). 
So if everything is ok with current PR then I'm able to fix DropdownMenu issue.

Thank you.

### DIFF
--- a/src/RootCloseWrapper.js
+++ b/src/RootCloseWrapper.js
@@ -86,7 +86,22 @@ export default class RootCloseWrapper extends React.Component {
   }
 
   componentDidMount() {
-    this.bindRootCloseHandlers();
+    if (!this.props.disabled) {
+      this.bindRootCloseHandlers();
+    }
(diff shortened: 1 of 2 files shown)


taion:
@jquense is this necessary given the `disabled` handling already present?

taion:
I ended up adding equivalent changes separately in https://github.com/react-bootstrap/react-overlays/pull/116. Sorry for the delay.

Issue #77, opened 2016-04-08
Include RootCloseWrapper as part of public API


cesarandreu:
`RootCloseWrapper` is small but definitely useful on its own. What are your thoughts about adding it as part of the public API?

jquense:
I'm a fan :+1:

taion:
Sounds good.

taion:
@cesarandreu Did you want to put together a PR?

leojh:
was anyone already working on this? if not, I'd be happy to try to do it this weekend.

taion:
I'm not aware that anyone is working on this.

andresgutgon:
I'm working on that ☝️

wzup:
Was the feature implemented or not?

taion:
No.

wzup:
@taion 

There are at least three files that import `RootCloseWrapper`
https://github.com/react-bootstrap/react-bootstrap/blob/a1a98485f1147654f244a201a3fe638b9a8333ec/src/DropdownMenu.js
https://github.com/react-bootstrap/react-overlays/blob/54a1fc47e09985a74d26a07982d94689a63cf46d/test/RootCloseWrapperSpec.js
https://github.com/react-bootstrap/react-overlays/blob/a268df0b60f3ec2b5afe5bb3456509274ba6f72c/src/Overlay.js

And here is a file with exactly the same component [RootCloseWrapper](https://github.com/react-bootstrap/react-overlays/blob/54a1fc47e09985a74d26a07982d94689a63cf46d/src/RootCloseWrapper.js)

So is it implemented? 

What's exactly is this component for? I need functionality that closes customized dropdown by clicking screen anywhere outside the dropdown. Is this the component that represents this functionality?

Morantron:
I read in #111 that docs are a blocker for this issue. What can we do to have this merged? Would commenting the code and adding some examples in `examples` folder be enough? Or would you need some examples in `gh-pages` branch as well?

aatrostle:
✋ I would love to help add docs to get this shipped, can I get a pointer to where to do that?

jquense:
the prop descriptions are generated from comments in the components and the docs are in the doc folder :thumbsup:

aatrostle:
_:gentle-nudge:_ #77 has been reviewed once and AFAIK should be good to go, let me know if theres anything else I need to address for it to ship.

PR #111, opened 2016-08-25
Expose rootCloseWrapper

andresgutgon:
## What is this?

We're doing our own popover component. And we want to accomplish 3 things with this component:
1. Be able to render outside of current React Element. Like in body. Done with `react-overlays/Portal`;
2. Close all popovers when clicking on the page. Done with `react-overlays/RootCloseWrrapper`
3. Position popover. Done with [popper.js](https://popper.js.org/)
## Why use Popper.js instead of Position component

Popper offer more placement options like `bottom-start`, `bottom-end`, `left-start`,...
Also has the concept of boundaries to react on scroll of an arbitrary element.

We think `react-overlays/Position` is fine but doesn't meet our requirements.
## It this posible?

It would be posible to expose in `RootCloseWrapper`? This way we can use a `Portal` and a `RootCloseWrapper` together which is an`Overlay` component. The problem is that `Overlay` component is using `react-overlays/Position` and we don't want that.

Thank you so much for this great package! :clap:

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -5,3 +5,4 @@ export Overlay from './Overlay';
 export Portal from './Portal';
 export Position from './Position';
 export Transition from './Transition';
+export RootCloseWrapper from './RootCloseWrapper';


taion:
This would close https://github.com/react-bootstrap/react-overlays/issues/77, but we'd need docs first.

taion:
I'm going to close this PR for the time being – the real blocker on #77 is adding documentation. The code here is simple enough that we can replicate it down the road as needed once we have those docs set up.

PR #113, opened 2016-09-23
Allow opt out of container styling, renderBackdrop

jquense:


src/Modal.js
```diff
@@ -212,7 +229,7 @@ const Modal = React.createClass({
       onExiting,
       onEnter,
       onEntering,
-      onEntered,
```
taion: o7 dangling comma

src/Modal.js
```diff
@@ -340,7 +365,7 @@ const Modal = React.createClass({
     let doc = ownerDocument(this);
     let container = getContainer(this.props.container, doc.body);
 
-    modalManager.add(this, container, this.props.containerClassName);
```
taion: what did this use to do?

src/ModalManager.js
```diff
@@ -20,21 +20,48 @@ function findContainer(data, modal) {
...
+    style.paddingRight =
+      parseInt(css(container, 'paddingRight') || 0, 10) + getScrollbarSize() + 'px';
+  }
+
+  css(container, style);
+}
+
+function removeContainerStyle({ style }, container) {
+
+  Object.keys(style).forEach(
```
taion: `Object.assign(container.style, style)`?

might need to pull in that Babel transform though

src/Modal.js
jquense: same thing, I just moved the default instance to a defaultProp so you can configure them via the component

jquense:
there is a bunch of clean up work we need to do here at somepoint
not sure waht the deal is with the Affix tests btw

PR #114, opened 2016-09-29
Allow accessibility attributes on the root element

adamschroder:
Hi there! I found some issues with this component in terms of accessibility. Particularly in how the Windows Narrator announces the Modal. From the [spec](https://www.w3.org/TR/wai-aria/roles#dialog) , the `aria-labelledby` or `aria-label` should be an attribute on the same element as the `role=dialog`.

Additionally, it would fit more use cases if the dialog could also be an [alertdialog](https://www.w3.org/TR/wai-aria/roles#alertdialog) (or anything else that could also make sense, semantically). 

If there are styling, formatting, or any other issues with this PR please let me know & i'd be happy to make them.


jquense:
Hi there, thanks for the PR. I think a more straightforward and flexible strategy here is to forward all the props (less the `Modal.propTypes`) to the component. Then there is no need for a whitelist.

src/Modal.js
```diff
@@ -267,20 +275,27 @@ const Modal = React.createClass({
...
-        <div
-          ref={'modal'}
-          role={role || 'dialog'}
-          style={style}
-          className={className}
-        >
-          { backdrop && this.renderBackdrop() }
-          { dialog }
-        </div>
+        {React.createElement('div',
```
jquense: Can you change this pack to jsx?

src/Modal.js
```diff
@@ -267,20 +275,27 @@ const Modal = React.createClass({
       );
     }
 
+    const modalProps = Object.assign(
```
jquense: instead of the whitelist you can do something like `elementProps = _.omit(this.props, Object.keys(Modal.PropTypes))`

test/ModalSpec.js
```diff
@@ -287,6 +287,41 @@ describe('Modal', function () {
...
+      <Modal aria-describedby="modal-description" show>
+        <strong id="modal-description">Message</strong>
+      </Modal>
+    , mountPoint);
+
+    let attr = instance.refs.modal.attributes.getNamedItem('aria-describedby').value;
+    expect(attr).to.equal('modal-description');
+  });
+
+  it('Should accept the `tabIndex` property on the Modal', function () {
```
jquense: I don't think it should accept a tab index here, that would screw up the focus management. We already ensure that the dialog component has a tabIndex for a11y

jquense:
Hmm there are a bunch of inconsistencies with how we manage roles here that I hadn't noticed before, in connection with the react-bootstrap modal. I need to think a bit on what might be the best way to move forward to fix them tho (pre-existing issues to your PR here)

adamschroder:
@jquense thanks for being so quick to review this PR. I've made your suggested changes, and I also took a stab at fixing some of the tests. Let me know of any other feedback.

adamschroder:
Not sure why the tests are failing on Travis, they are passing locally. Any ideas?

src/Modal.js
```diff
@@ -267,6 +268,15 @@ const Modal = React.createClass({
       );
     }
 
+    const keys = Object.keys(_props);
+    const propTypes = Object.keys(Modal.propTypes);
+    const newProps = {};
+    keys.map(function (prop) {
+      if (!propTypes.includes(prop)) {
```
taion: `Array.prototype.includes` is not available in all browsers. That's why tests are failing. Separately this should be an arrow function though. Even more separately this should just be what @jquense said with `_.omit`.

adamschroder: Hey @taion, great catch. I didn't use `_.omit` because lodash is just a dev dependency in this project, and I didn't want to add it just for this. If ya'll are ok with adding it, i'll happily change the code.

taion:
@jquense Are your concerns regarding the pre-existing issues blockers for this PR?

adamschroder:
Hey guys, whats the status on this PR?

adamschroder:
hey @jquense, it has been a little over 2 weeks since there was any feedback here, any updates?

src/Modal.js
```diff
@@ -267,6 +268,15 @@ const Modal = React.createClass({
...
     }
 
+    const keys = Object.keys(_props);
+    const propTypes = Object.keys(Modal.propTypes);
+    const newProps = {};
+    keys.map((prop) => {
+      if (propTypes.indexOf(prop) === -1) {
+        newProps[prop] = _props[prop];
+      }
+    });
```
jquense: can you hoist this out of hte render method as an `omitProps` function just so its a bit clearer what's going on?

jquense:
i guess this is ok as is for now, could you also rebase the branch against master so it's a bit cleaner

thanks!

adamschroder:
Updated & rebased. Thanks for the attention and feedback!

taion:
ping @jquense

adamschroder:
Hey guys, sorry for the spam, but can we try to get this merged?

PR #116, opened 2016-09-30
Clean up RootCloseWrapper

taion:


src/Portal.js
```diff
@@ -102,11 +102,7 @@ let Portal = React.createClass({
     }
 
     if (this._overlayInstance) {
-      if (this._overlayInstance.getWrappedDOMNode) {
-        return this._overlayInstance.getWrappedDOMNode();
```
jquense: what was this even for?

taion: Just compat with the old RCW... this method was only defined on RCW and doing a quick GH code search, the only instances of it are people vendoring this code

src/RootCloseWrapper.js
```diff
@@ -85,62 +61,44 @@ export default class RootCloseWrapper extends React.Component {
...
-        {child}
-      </div>
-    );
+    if (this.props.onRootClose) {
+      this.props.onRootClose();
+    }
   }
 
-  getWrappedDOMNode() {
-    // We can't use a ref to identify the wrapped child, since we might be
```
jquense: ahh right

src/RootCloseWrapper.js
```diff
@@ -18,31 +13,32 @@ function isModifiedEvent(event) {
...
 
-    let { id, suppressRootClose } = getSuppressRootClose();
+  componentDidMount() {
+    if (!this.props.disabled) {
+      this.bindRootCloseHandlers();
+    }
+  }
 
-    this._suppressRootId = id;
+  componentDidUpdate(prevProps) {
```
jquense: is there a reason to do this here vs directly in cWRP?

taion: no... do we generally do event binding stuff in cWRP instead of in cDU though?

jquense: generally cDU. this is fine, i don't think there is any reason why a child component would need the handlers registered before it rendered/updated.

jquense:
lgtm

src/RootCloseWrapper.js
taion: yeah that's why i put it here... thought we usually did this kind of thing in cDU (though i agree here it could be in either place)

PR #117, opened 2016-09-30
Fix rootClose on unmounting components

taion:
Most likely React unmounted the target in response to some other click.

@jquense This logic is not ideal but probably fine in practice? What do you think?

### DIFF
--- a/src/RootCloseWrapper.js
+++ b/src/RootCloseWrapper.js
@@ -44,8 +44,11 @@ export default class RootCloseWrapper extends React.Component {
   bindRootCloseHandlers() {
     const doc = ownerDocument(this);
 
+    // Use capture for this listener so it fires before React's listener, to
+    // avoid false positives in the contains() check below if the target DOM
+    // element is removed in the React mouse callback.
     this._onDocumentMouseListener =
-      addEventListener(doc, this.props.event, this.handleDocumentMouse);
+      addEventListener(doc, this.props.event, this.handleDocumentMouse, true);
 
     this._onDocumentKeyupListener =
       addEventListener(doc, 'keyup', this.handleDocumentKeyUp);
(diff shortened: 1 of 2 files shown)


jquense:
par* why can't I edit anything

taion:
The alternative is to go back to the old approach, which ought to be safe to this kind of thing. wdyt? I think the "replace popover" scenario is rare enough that we don't have to rush out another release.

taion:
Actually, I don't think this works. Imagine the user clicks on a router link and navigates... this logic would still make the page transition. Is there a way to attach our document listener to fire _before_ the React click handler?

taion:
@jquense You know DOM stuff a lot better than me... this is a reasonable fix?

jquense:
capture seems fine...the thought is that if we fire before react, then there isn't any chance of the target being nonexistent due to an unmount caused by the event? I think its reasonable, I guess technically you could break it by using `onClickCapture` on a child component, but I think there are like 4 people that know that api exists in React, so probably not a problem :P.

taion:
What the heck is onClickCapture? What even _is_ capture supposed to be other than an escape hatch for priority?

jquense:
there aren't a ton of use cases for the capture flag, but generally its used for event delegation, e.g. you can mimic `focusin` event behavior by using `focus` with capture set to `true`.

The react event system will let you listen for the events with `capture` `true` if you add the "Capture" suffix to the event name(for events that support capture that is). I don't think anyone uses it but you can do like: `<input onChangeCapture={handler}/>` , At least you used to be able to.

taion:
I don't understand what any of that means. But, yeah, in practice this seems to mean that we fire before the target could potentially get unmounted by a normal React `onClick` or whatever handler.

jquense:
:rainbow:

PR #118, opened 2016-10-02
Actually fix RootCloseWrapper

taion:
I'm not sure how I messed up checking R-B tests. We need to defer firing the callback until the actual bubble phase, because otherwise the toggle functionality in `<OverlayTrigger>` will toggle the overlay back _open_ if a user clicks on the button. We actually have a test case for this – I guess I just didn't check it or something before releasing v0.6.9.


jquense:
software is hard.

PR #124, opened 2016-10-28
Fix whack github highlighting

camjc:
Should be using a curly apostrophe anyway, so two birds with one stone here.

### DIFF
--- a/examples/Overlay.js
+++ b/examples/Overlay.js
@@ -113,7 +113,7 @@ const OverlayExample = React.createClass({
           target={ props => findDOMNode(this.refs.target)}
         >
           <ToolTip>
-            I’m placed to the: <strong>{this.state.placement}</strong>
+            I&rsquo;m placed to the: <strong>{this.state.placement}</strong>
           </ToolTip>
         </Overlay>
       </div>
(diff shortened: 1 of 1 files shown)


taion:
I don't think we want non-ASCII source

camjc:
No worries, shall I close this?

taion:
Depends how much this syntax highlighting thing bothers you. The HTML entity for an appropriate apostrophe could work.

taion:
Thanks!

examples/Overlay.js
```diff
@@ -113,7 +113,7 @@ const OverlayExample = React.createClass({
           target={ props => findDOMNode(this.refs.target)}
         >
           <ToolTip>
-            I'm placed to the: <strong>{this.state.placement}</strong>
```
oliviertassinari: I believe that we could have solve it this way:
```jsx
{"I'm placed to the: <strong>{this.state.placement}</strong>"}
```
To be confirmed.

taion: Yeah... frankly I'm surprised the GH syntax highlighting fails so badly here. Oh well.

taion: This isn't actually the JSX mode per se – there is a GH JSX mode but it's a bit wonky.

PR #138, opened 2017-01-05
Update dom-helpers dependency to include requestAnimationFrame fix.

ruiaraujo:


taion:
This is sort of bizarre... we never cancel RAF callbacks here, do we?

ruiaraujo:
@taion I do, and I did not want to include two version of dom-helpers. 😄

PR #142, opened 2017-01-14
Resolves #141 Expose DOM event for close events from the RootCloseWrapper.

### LINKED ISSUE #141: How to tell the difference between click and esc close?
With `<Overlay />` Is there a way to tell the difference of how the close callback got called? Specifically between a click and pressing the esc key?

https://github.com/react-bootstrap/react-overlays/blob/master/src/RootCloseWrapper.js#L84
https://github.com/react-bootstrap/react-overlays/blob/master/src/RootCloseWrapper.js#L90

It seems odd to me that nothing is passed to the callback. No string, nor the DOM event to work with.

For my use case
* Pressing `esc` means cancel & close.
* Clicking on background means save & close.

TheBox193:
Expose the DOM event through the callback. Allows the callback to determine how the RootCloseWrapper was closed, by `click` or by `keyup` of `esc`.

See #141

### DIFF
--- a/src/RootCloseWrapper.js
+++ b/src/RootCloseWrapper.js
@@ -79,15 +79,15 @@ export default class RootCloseWrapper extends React.Component {
     );
   };
 
-  handleMouse = () => {
+  handleMouse = (e) => {
     if (!this.preventMouseRootClose && this.props.onRootClose) {
-      this.props.onRootClose();
+      this.props.onRootClose(e);
     }
   };
 
   handleKeyUp = (e) => {
     if (e.keyCode === 27 && this.props.onRootClose) {
-      this.props.onRootClose();
+      this.props.onRootClose(e);
     }
   };
 
--- a/src/RootCloseWrapper.js
+++ b/src/RootCloseWrapper.js
@@ -5,6 +5,8 @@ import ReactDOM from 'react-dom';
 import addEventListener from './utils/addEventListener';
 import ownerDocument from './utils/ownerDocument';
 
+const escapeKeyCode = 27;
+
 function isLeftClickEvent(event) {
   return event.button === 0;
 }
@@ -86,7 +88,7 @@ export default class RootCloseWrapper extends React.Component {
   };
 
   handleKeyUp = (e) => {
-    if (e.keyCode === 27 && this.props.onRootClose) {
+    if (e.keyCode === escapeKeyCode && this.props.onRootClose) {
       this.props.onRootClose(e);
     }
   };
--- a/test/RootCloseWrapperSpec.js
+++ b/test/RootCloseWrapperSpec.js
@@ -4,6 +4,8 @@ import RootCloseWrapper from '../src/RootCloseWrapper';
 import { render } from './helpers';
 import simulant from 'simulant';
 
+const escapeKeyCode = 27;
+
 describe('RootCloseWrapper', function () {
   let mountPoint;
 
@@ -29,10 +31,6 @@ describe('RootCloseWrapper', function () {
     shouldCloseOn('mousedown', 'mousedown');
   });
 
-  describe('using keyup event', () => {
-    shouldCloseOnKeyUp('keyup');
-  });
-
   function shouldCloseOn(eventProp, eventName) {
     it('should close when clicked outside', () => {
       let spy = sinon.spy();
@@ -50,7 +48,7 @@ describe('RootCloseWrapper', function () {
 
       expect(spy).to.have.been.calledOnce;
 
-      assert.oneOf(spy.getCall(0).args[0].type, ['click', 'mousedown'] );
+      expect(innerSpy.getCall(0).args[0].type).to.be.oneOf(['click', 'mousedown']);
     });
 
     it('should not close when right-clicked outside', () => {
@@ -107,11 +105,11 @@ describe('RootCloseWrapper', function () {
       expect(outerSpy).to.have.not.been.called;
       expect(innerSpy).to.have.been.calledOnce;
 
-      assert.oneOf(innerSpy.getCall(0).args[0].type, ['click', 'mousedown']  );
+      expect(innerSpy.getCall(0).args[0].type).to.be.oneOf(['click', 'mousedown'])
     });
   }
 
-  function shouldCloseOnKeyUp(eventName) {
+  describe('using keyup event', () => {
     it('should close when escape keyup', () => {
       let spy = sinon.spy();
       render(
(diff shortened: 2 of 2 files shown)


src/RootCloseWrapper.js
```diff
@@ -81,13 +81,13 @@ export default class RootCloseWrapper extends React.Component {
 
   handleMouse = () => {
```
chrisege: I do you need to pass `e` here too?

src/RootCloseWrapper.js
```diff
@@ -81,13 +81,13 @@ export default class RootCloseWrapper extends React.Component {
...
     if (!this.preventMouseRootClose && this.props.onRootClose) {
-      this.props.onRootClose();
+      this.props.onRootClose(e);
```
chrisege: Otherwise I think this `e` (Not defined) might cause an error.

TheBox193:
@jquense Updated.

taion:
Please add a test here

taion:
(just to prevent future regressions)

codebreach:
@TheBox193 any chance you will get around to that test anytime soon? If you can't, I can push a commit with the tests.
This is blocking our release (via https://github.com/react-bootstrap/react-bootstrap/pull/2422)

taion:
We'd be happy to accept a PR that incorporates this code and adds a test – just clone the branch and add a new commit.

TheBox193:
Any thoughts how to verify the args are an event? I presume something along the lines of `expect(spy.args)`. But I'm not sure how to validate the arg is a proper click event / key event.

If someone understands sinon better than me feel free to add a commit / add pr to my branch.

jquense:
if you pass a `sinon.spy()` to as a handler you can check the args like `spy.getCall(0).args[0]` and just assert the event type is correct or something.

codebreach:
See #143

test/RootCloseWrapperSpec.js
```diff
@@ -100,6 +106,53 @@ describe('RootCloseWrapper', function () {
...
+  //                 <div id='my-other-div'>hello there</div>
+  //               </RootCloseWrapper>
+  //             </div>
+  //           </RootCloseWrapper>
+  //         );
+
+  //         simulant.fire(document.body, eventName, {keyCode: 27});
+
+  //         expect(outerSpy).to.have.not.been.called;
+  //         expect(innerSpy).to.have.been.calledOnce;
```
TheBox193: It appears that an `esc` keyup event is causing both `onRootClose` callbacks to be called, thus this test will currently fail. We should open a separate issue to address that to keep this PRs scope smaller.

jquense: I'd really prefer the the test **not** be commented out. If the PR is exposing a bug then we should fix it. At the very least the test should be left as it was, vs commented out completely.

src/RootCloseWrapper.js
```diff
@@ -79,15 +79,15 @@ export default class RootCloseWrapper extends React.Component {
...
-  handleMouse = () => {
+  handleMouse = (e) => {
     if (!this.preventMouseRootClose && this.props.onRootClose) {
-      this.props.onRootClose();
+      this.props.onRootClose(e);
     }
   };
 
   handleKeyUp = (e) => {
     if (e.keyCode === 27 && this.props.onRootClose) {
```
rringler: I realize this didn't originate from your PR, but is it worth extracting `27` into a local that gives some context (eg. `escapeKeyCode`)?

test/RootCloseWrapperSpec.js
```diff
@@ -100,6 +106,53 @@ describe('RootCloseWrapper', function () {
...
+      let spy = sinon.spy();
+      render(
+        <RootCloseWrapper onRootClose={spy}>
+          <div id='my-div'>hello there</div>
+        </RootCloseWrapper>
+      );
+
+      expect(spy).to.not.have.been.called;
+
+      simulant.fire(document.body, eventName, {keyCode: 27});
```
rringler: Maybe here too?  Either a local (or even just a comment that explains `27` is the keyCode for escape)?

test/RootCloseWrapperSpec.js
TheBox193: Agreed it should be fixed since this test exposes and an existing issue.

I'm afraid I don't have the time to commit to it. Since it is an existing bug I would say either open an issue where it can be addressed or someone willing to champion the scope change of this PR should take it on.

jquense: that's fine if you don't want to fix it here, I understand you might not have time, but please put the test back to the way it was before your PR, in stead of commenting it out.

TheBox193: It's a new test I wrote. Nothing removed, only added, and hopefully we could turn it on someday.

jquense: ok sorry, the way you added these tests is a bit confusing.

test/RootCloseWrapperSpec.js
```diff
@@ -100,6 +106,53 @@ describe('RootCloseWrapper', function () {
 
       expect(outerSpy).to.have.not.been.called;
       expect(innerSpy).to.have.been.calledOnce;
+
+      assert.oneOf(innerSpy.getCall(0).args[0].type, ['click', 'mousedown']  );
+    });
+  }
+
+  function shouldCloseOnKeyUp(eventName) {
```
jquense: this shouldn't be a function, since you only call it once. Instead inline the tests below in the keydown describe block

test/RootCloseWrapperSpec.js
```diff
@@ -45,6 +49,8 @@ describe('RootCloseWrapper', function () {
       simulant.fire(document.body, eventName);
 
       expect(spy).to.have.been.calledOnce;
+
+      assert.oneOf(spy.getCall(0).args[0].type, ['click', 'mousedown'] );
```
taion: don't mix and match assertion styles within a single test

test/RootCloseWrapperSpec.js
taion: this behavior isn't obviously wrong to me anyway, and i don't want to faff around a root close wrapper manager or something like that

test/RootCloseWrapperSpec.js
```diff
@@ -100,6 +104,57 @@ describe('RootCloseWrapper', function () {
...
+      );
+
+      expect(spy).to.not.have.been.called;
+
+      simulant.fire(document.body, 'keyup', {keyCode: escapeKeyCode});
+
+      expect(spy).to.have.been.calledOnce;
+
+      expect(spy.getCall(0).args.length).to.be.eql(1);
+      expect(spy.getCall(0).args[0]).to.not.be.undefined;
```
taion: these assertions are pointless if you're already asserting on the event below

test/RootCloseWrapperSpec.js
```diff
@@ -100,6 +104,57 @@ describe('RootCloseWrapper', function () {
...
+        </RootCloseWrapper>
+      );
+
+      expect(spy).to.not.have.been.called;
+
+      simulant.fire(document.body, 'keyup', {keyCode: escapeKeyCode});
+
+      expect(spy).to.have.been.calledOnce;
+
+      expect(spy.getCall(0).args.length).to.be.eql(1);
```
taion: these should be `.to.equal`

test/RootCloseWrapperSpec.js
```diff
@@ -100,6 +104,57 @@ describe('RootCloseWrapper', function () {
...
+            <RootCloseWrapper onRootClose={innerSpy}>
+              <div id='my-other-div'>hello there</div>
+            </RootCloseWrapper>
+          </div>
+        </RootCloseWrapper>
+      );
+
+      simulant.fire(document.body, 'keyup', {keyCode: escapeKeyCode});
+
+      // TODO: Update library to make this expectation pass.
```
taion: not sure this TODO is correct

taion:
@jquense You okay with merging this?

chrisege:
🎉

PR #143, opened 2017-02-07
Resolves #141 Expose DOM event for close events from the RootCloseWrapper

### LINKED ISSUE #141: How to tell the difference between click and esc close?

codebreach:
Taken form #142 
Will unblock https://github.com/react-bootstrap/react-bootstrap/pull/2422

### DIFF
--- a/src/RootCloseWrapper.js
+++ b/src/RootCloseWrapper.js
@@ -79,15 +79,15 @@ export default class RootCloseWrapper extends React.Component {
     );
   };
 
-  handleMouse = () => {
+  handleMouse = (e) => {
     if (!this.preventMouseRootClose && this.props.onRootClose) {
-      this.props.onRootClose();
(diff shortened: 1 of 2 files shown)


TheBox193:
Added tests in #142, they might be more comprehensive. I'm fine with either.

taion:
Closing this in favor of iterating on #142.

PR #144, opened 2017-02-15
add mountOnEnter

jquense:
Solves the dang lazy tab loading thing downstream, at least for animation.

also cleans up the logic a bit (I think please double check)

### DIFF
--- a/src/Transition.js
+++ b/src/Transition.js
@@ -25,12 +25,14 @@ class Transition extends React.Component {
   constructor(props, context) {
     super(props, context);
 
+    const { transitionAppear, unmountOnExit, mountOnEnter } = props;
+
     let initialStatus;
     if (props.in) {
       // Start enter transition in componentDidMount.
-      initialStatus = props.transitionAppear ? EXITED : ENTERED;
+      initialStatus = transitionAppear ? EXITED : ENTERED;
     } else {
-      initialStatus = props.unmountOnExit ? UNMOUNTED : EXITED;
+      initialStatus = unmountOnExit || mountOnEnter ? UNMOUNTED : EXITED;
     }
     this.state = {status: initialStatus};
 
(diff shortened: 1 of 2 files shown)


jquense:
🏏

src/Transition.js
```diff
@@ -44,51 +46,41 @@ class Transition extends React.Component {
...
+    this._propsUpdated = true
+
+    if (nextProps.in && this.state.status === UNMOUNTED) {
+      // Start enter transition in componentDidUpdate.
+      this.setState({ status: EXITED });
     }
   }
 
   componentDidUpdate() {
     const status = this.state.status;
```
taion: surprised we don't destructure here

src/Transition.js
```diff
@@ -44,51 +46,41 @@ class Transition extends React.Component {
...
-      // EXITED is always a transitional state to either ENTERING or UNMOUNTED
-      // when using unmountOnExit.
-      if (this.props.in) {
-        this.performEnter(this.props);
-      } else {
-        this.setState({status: UNMOUNTED});
-      }
+    // EXITED is always a transitional state to either ENTERING or UNMOUNTED
+    // when using unmountOnExit.
+    const needsTransitionalUpdate = (
```
taion: this is a little confusing because a transitional update is needed when `mountOnEnter` and `in` are both set

src/Transition.js
```diff
@@ -44,51 +46,41 @@ class Transition extends React.Component {
...
-        this.setState({status: EXITED});
-      }
-    }
-    else {
-      this._needsUpdate = true
+    this._propsUpdated = true
+
+    if (nextProps.in && this.state.status === UNMOUNTED) {
+      // Start enter transition in componentDidUpdate.
+      this.setState({ status: EXITED });
```
taion: just for clarity i liked not setting `_propsUpdated` on this path, since the transitional update logic will handle this

src/Transition.js
```diff
@@ -44,51 +46,41 @@ class Transition extends React.Component {
...
-        }
-        // Otherwise we're already exited or exiting.
+    this._propsUpdated = false;
+
+    if (this.props.in) {
+      if (status === EXITING || status === EXITED) {
+        this.performEnter(this.props);
       }
     }
+    else if (needsTransitionalUpdate) {
```
taion: ahh... maybe this would be better named `needsUnmountForExit` or something?

taion: or maybe `needsTransitionalUpdate && unmountOnExit`

taion:
Thinking about this a bit more – can we just have a `nextStatuses` array that we manage purely in cWRP instead of splitting between cWRP and cDU, then just read from that in cDU?

I think I was overly perf focused when I wrote this initially, and that doing so was the wrong call.

jquense:
Some of the reason this is in CDU is because it needs to respond to transitional changes that result from `state` changes, as well as `prop` ones. That said I do think there is probably a simpler logic to be had here, but I was too afraid of odd edge cases cropping up that I didn't go the full rewrite path :P 

I'll be the first to admit i'm not super happy with the awkward logic branches here.

src/Transition.js
```diff
@@ -18,109 +18,101 @@ export const EXITING = 4;
...
 
   componentDidMount() {
-    if (this.props.transitionAppear && this.props.in) {
-      this.performEnter(this.props);
-    }
+    this.updateStatus();
   }
 
   componentWillReceiveProps(nextProps) {
-    this._propsUpdated = true
```
taion: wait why do we always defer rendering `ENTERING` until the next render cycle?

taion: oh i see, there's a test asserting so

taion:
@jquense wdyt?

jquense:
Looks good to me, from what I could tell. I haven't had a moment to try it out tho. IF you feel good about it :shipit: tho

taion:
carpe yolo

PR #145, opened 2017-02-15
Modal: add restoreFocus option

erykpiast:
Allow to specify if last focus should be restored after modal is hidden.

I suppose it may be possible and appreciated to write test for it, like those ones https://github.com/react-bootstrap/react-overlays/blob/master/test/ModalSpec.js#L312, but before I spend a lot of time on it, I'd like to be sure you're OK with this change.

The thing is, it's not always desired to call `focus` method on thing that was `activeElement` at the moment modal was shown. I can imagine it's valid in most of use cases, however in some it's not and... I did things I'm really, really ashamed of to workaround this. Please, let my conscience to be cleared.

### DIFF
--- a/src/Modal.js
+++ b/src/Modal.js
@@ -158,6 +158,12 @@ const Modal = React.createClass({
      * accessible to assistive technologies, like screen readers.
      */
     enforceFocus: React.PropTypes.bool,
+    
+    /**
+     * When `true` The modal will restore focus to previously focused element once
+     * modal is hidden
+     */
+    restoreFocus: React.PropTypes.bool,
 
     /**
      * Callback fired before the Modal transitions in
@@ -205,6 +211,7 @@ const Modal = React.createClass({
       keyboard: true,
       autoFocus: true,
       enforceFocus: true,
+      restoreFocus: true,
       onHide: noop,
       manager: modalManager,
       renderBackdrop: (props) => <div {...props} />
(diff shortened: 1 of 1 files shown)


jquense:
LGTM

taion:
Should we check for something like `autoFocus && restoreFocus`? Seems like it'd be odd to restore focus if `autoFocus` is unset. Alternatively, do we even need `restoreFocus` as separate from `autoFocus`?

erykpiast:
So far It was working independently from each other.

taion:
Should we restore the focus if `autoFocus` is unset?

erykpiast:
So far you was doing it :P I really think those are separate features. The thing is, that one is optional and another is not. As simple as that. You'll never cover all the use cases of different people, not in such library as Bootstrap. Moreover, changing default behavior may be treated as backward incompatible API change I suppose, so I'm not sure if it's worth it right now. Probably is worth to create GH issue and discuss the problem there.

taion:
@jquense What do you think? Should we restore focus by default if the user only disables `autoFocus`?

taion:
@jquense ping

jquense:
I feel fine with them all being seperate toggles, they are really last ditch escape hatches anyway, so it seems ok to mix and match as you please. Most shouldn't turn any of them off

PR #149, opened 2017-03-02
Expose RootCloseWrapper as part of public API

### LINKED ISSUE #77: Include RootCloseWrapper as part of public API

aatrostle:
This PR exposes `<RootCloseWrapper/>` as part of the public API and adds examples. Resolves #77.

Thanks to the maintainers, this package has been super helpful ❤️

### DIFF
--- a/src/RootCloseWrapper.js
+++ b/src/RootCloseWrapper.js
@@ -17,7 +17,8 @@ function isModifiedEvent(event) {
 
 /**
  * The `<RootCloseWrapper/>` registers your callback on the document when rendered. Powers the
- * `<Overlay/>` component.
+ * `<Overlay/>` component. This is used achieve modal style behavior where your callback is
+ * triggered when the user tries to interact with the rest of the document or hits the `esc` key.
  */
 class RootCloseWrapper extends React.Component {
   constructor(props, context) {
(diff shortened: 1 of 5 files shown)


src/RootCloseWrapper.js
```diff
@@ -15,7 +15,11 @@ function isModifiedEvent(event) {
   return !!(event.metaKey || event.altKey || event.ctrlKey || event.shiftKey);
 }
 
-export default class RootCloseWrapper extends React.Component {
+/**
+ * The `<RootCloseWrapper/>` registers your callback on the document when rendered. Powers the
```
taion: worth noting this also triggers on users hitting `esc`

aatrostle: Should I note that here in the desc or in the `onRootClose` prop?

taion: possibly both – and also describe the purpose of root close wrapper (i.e. to make popovers behave in a modal manner where they close when the user tries to interact with the rest of the doucment)

aatrostle:
@taion Review feedback addressed 👍

jquense:
LGTM

PR #150, opened 2017-03-05
Portal: Move _renderOverlay() to component render

Vija02:
As far as I can see, there's no reason to use componentDidMount and componentDidUpdate to render the overlay. Might as well render it on the render function.

Also, this will make it not reliant on the order of the lifecycle which will avoid problems like [developit/preact-compat#312](https://github.com/developit/preact-compat/issues/312)

### DIFF
--- a/src/Portal.js
+++ b/src/Portal.js
@@ -24,14 +24,6 @@ let Portal = React.createClass({
     ])
   },
 
-  componentDidMount() {
-    this._renderOverlay();
-  },
-
-  componentDidUpdate() {
-    this._renderOverlay();
-  },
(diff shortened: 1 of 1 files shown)


taion:
That's not correct. This is intentional.

jquense:
render in general shouldn't have side effects, but also from a code viewpoint this won't work on the first render in where the component isn't mounted yet

Vija02:
Care to explain why this won't work on first render and why it's intentional :smile: ? I tried some simple tests and everything seems smooth. Just curious here hehe. Thanks!

PR #154, opened 2017-03-09
Add RootCloseWrapper's event support to Overlay

guillaume86:
Just adds support for RootCloseWrapper's "event" prop in the Overlay component.


src/Overlay.js
```diff
@@ -77,8 +77,13 @@ class Overlay extends React.Component {
 
     // This goes after everything else because it adds a wrapping div.
     if (rootClose) {
+      let { event } = props;
```
taion: eh why use a different pattern here v for pulling out `onHide` below?

PR #155, opened 2017-03-10
Import Transition into Modal to fix default property value

skuridin:
[Here](https://github.com/react-bootstrap/react-overlays/blob/master/src/Modal.js#L243) `<Modal />` tries to fallback to `<Transition />`, but it was never imported. So it works like: `prop = undefined || undefined`. Because of this, `renderBackdrop()` will never work without passing `transition` property to  the `<Modal />`.

Probably proper way will be to put `<Transition />` to `getDefaultProps()`, but I think it's better to change as less code as possible.

### DIFF
--- a/src/Modal.js
+++ b/src/Modal.js
@@ -6,6 +6,7 @@ import elementType from 'react-prop-types/lib/elementType';
 
 import Portal from './Portal';
 import ModalManager from './ModalManager';
+import Transition from './Transition';
 
 import ownerDocument from './utils/ownerDocument';
 import addEventListener from './utils/addEventListener';


taion:
that's not what that code does

osener:
@taion Can you elaborate on this? Using the latest version of `react-overlay`, we've found that the function passed as the value of `renderBackdrop` prop doesn't get called at all unless you also pass `transition` property to `Modal`.

Please check out this example: https://www.webpackbin.com/bins/-Kf7B2HDFFEQ3H0MXYwk

taion:
that `transition: Transition` bit is equivalent to saying `const Transition = this.props.transition` and has no bearing on the actual exported `Transition` module

jquense:
I think the underlying issue is that renderBackdrop (for whatever reason) doesn't get called when there is no transition. That problem tho isn't correctly solved by making Transition defined in that method

PR #159, opened 2017-04-04
Use a function to store the modal node reference

jochenberger:


jochenberger:
This probably requires a minor version bump.

taion:
code looks fine. @jquense thoughts on versioning strategy here?

jquense:
it'd be a "major" bump to be safe. folks rely on the refs for testing. Need to see if RB would break as well actually

taion:
we're on minor versions :p LGTM

jquense:
hence the scare quotes :P

jochenberger:
If you think that people need access to the `div`, then maybe we should add a `modalRef` prop to the `Modal`.

taion:
did this get pulled in with that other PR?

jochenberger:
Oh, it seems I have accidentally deleted the branch. Will reopen.

PR #160, opened 2017-04-04
Consistently use renderBackdrop

jquense:
Not sure why this was this way. Now it uses `renderBackdrop` even without a transition

### DIFF
--- a/src/Modal.js
+++ b/src/Modal.js
@@ -317,14 +317,12 @@ const Modal = React.createClass({
 
     const backdropRef = ref => this.backdrop = ref;
 
-    let backdrop = (
-      <div
-        ref={backdropRef}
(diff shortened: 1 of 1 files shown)


src/Modal.js
```diff
@@ -317,27 +317,20 @@ const Modal = React.createClass({
...
 
-    let backdrop = (
-      <div
-        ref={backdropRef}
-        style={this.props.backdropStyle}
-        className={this.props.backdropClassName}
-        onClick={this.handleBackdropClick}
-      />
-    );
+    let backdrop = renderBackdrop({
```
taion: const? :p

jquense: its redefined below ;)

PR #163, opened 2017-04-08
Replace createClass with ES6 classes

JeffreyATW:
Changes all `createClass` calls to ES6 classes. Many instance methods are now arrow functions to maintain proper context.

Also updated `react-bootstrap` to reduce (not eliminate) warnings within the examples page.
That required an update to `sinon`.

Also culled React 15.5 `PropTypes` warnings by pulling in `prop-types` and referring to them via a codemod.


jquense:
thanks! I'll take a look when I get a chance

taion:
Quick initial comment is that while it's semantics-preserving to convert all the methods to bound instance methods, doing so isn't how one would "naturally" write a class component, and that it'd be better of the ones that weren't bound handlers weren't written as class property arrow functions.

JeffreyATW:
I didn't convert all of them to bound handlers - only the ones using `this`. I can take a closer look at how each method is called, but that was my general rule.

taion:
Did you do this by hand or with a code mod? Nothing here's strictly wrong, it's just that if you were writing something like this from scratch, you'd only bind the handlers instead of all the methods.

JeffreyATW:
I did the binding by hand. I was going a bit too fast to determine which instance methods were handlers and which ones were not. I could undo the changes to the methods that aren't used as handlers if you think it's worth it.

taion:
Probably not worth it. Just something to leave as a low-priority FIXME for later.

JeffreyATW:
I'll be entirely honest - I didn't even notice the `createClass` codemod at the bottom of [this page](https://github.com/reactjs/react-codemod), which is why I only used a codemod for the `PropTypes` change. I think I essentially achieved the same result, but would be interested in seeing how similar the codemod's result is...

JeffreyATW:
I tried the codemod and it didn't work. Would have had to deal with deprecated methods like `isMounted` anyway. Never mind that!

sorahn:
@taion this is unrelated to this PR, but it's something I've been wondering.  Why *not* just leave all the methods as fat arrow syntax now? Is there any language related reasons not to do that, or is it just preference.

Pure curiosity at this point.  I tested it in a sample app and even changing things like `render` and `componentWillMount` to fat arrows seemed to work just fine, so why not do it?

taion:
There should be a marginal performance benefit – no need to create the function as a bound function every time we instantiate the class.

PR #166, opened 2017-04-19
Use prop-types instead of React.PropTypes

jochenberger:
`React.PropTypes` is deprecated, `prop-types` should be used instead.
https://facebook.github.io/react/blog/2017/04/07/react-v15.5.0.html
https://www.npmjs.com/package/prop-types


SavePointSam:
Can this update be pushed out as a minor release? Would like like to update and clear out my console warnings. Thanks!

PR #168, opened 2017-04-20
Remove duplicate import statements

aatrostle:
While upgrading for 15.5 I noticed this package had been recently updated but the latest build was failing.

Fixes duplicate `prop-types` import statements in multiple files.


jquense:
Thanks!

Issue #135, opened 2016-12-19
margin on Position target causes incorrect positioning


timothykang:
This issue can be reproduced directly on [the `Position` example](https://react-bootstrap.github.io/react-overlays/#position), by changing
```jsx
<Button bsStyle='primary' ref='target' onClick={this.toggle}>
```
to
```jsx
<Button bsStyle='primary' ref='target' onClick={this.toggle} style={{marginLeft: '10px'}}>
```

This issue looks to be caused by the usage of `dom-helpers/query/position` (in `utils/calculatePosition.js`). `dom-helpers/query/position` mimics [jQuery.position(), which does not account for margins](https://api.jquery.com/position/).

I believe the correct course of action is to account for margins (i.e. add them to the offsets) in `utils/calculatePosition.js`. If the maintainers agree I'd be happy submit a PR.

taion:
What does upstream TWBS do?

jquense:
jquery may include margins in dimension calculations... I'll have to check

timothykang:
It looks like TWBS uses jQuery.offset() when getting and setting positions (see [js/tooltip.js](https://github.com/twbs/bootstrap/blob/v3.3.7/js/tooltip.js), lines 295 + 370).

devongovett:
See #175 for a fix.

taion:
We actually can't do this without breaking Bootstrap compatibility. Bootstrap popovers actually do apply margins and depend on them for the positioning to look right.

This is sort of why we need to pull `<Position>` as-is into React-Bootstrap – it's just not in general a great primitive for anything other than TWBS overlays.

qrohlf:
@taion unless I'm mistaken, popover margins are a seperate issue from this ticket which is talking about target element margins.

I've taken a stab at fixing the issue with target element margins only, which you can see here: https://github.com/react-bootstrap/react-overlays/pull/206

Open question: should overlays be positioned inside or outside an element's margins by default? I can see a case for both options.

Issue #52, opened 2015-12-12
More placement types for Position


vladimir-rovensky:
Would it be possible to add more placement types for the `Position` component? It works great, but I can only position in one dimension, the other gets centered (e.g. placement=bottom means essentially bottom center).

Ideally, the `Position` component would have a similar API to that of [jQuery UI Position](https://jqueryui.com/position/) - the user says what point of the element should be positioned to what point of the target. Is that realistic in the context of the `Position` component?  .

jquense:
I think its reasonable for this component, however the major use case for it has been bootstrap tooltips,  which is mostly covered by the more simplistic placements. 

Which is to say that I am not sure anyone on the team haa the time or the use-case for adding more placements, but we'd happily accept a PR ;)

bluetidepro:
@jquense @vladimir-rovensky Did you ever figure out a good way to handle this? Also looking to use positioning a bit more advance like `top left`, `bottom right`, etc. It's very useful to get the carrot in the right position for tooltips.

vladimir-rovensky:
Not really, I ended up using [jquery-ui position](https://jqueryui.com/position/), it was the quickest solution I could think of, plus it handles collisions. Would be nice to have a pure react solution for this though.

bluetidepro:
@vladimir-rovensky Do you have an example of how you did that and use jquery-ui positioning to do it?

vladimir-rovensky:
I made a simple wrapper component kinda like this:

```
React.createClass({
    render(){
        return <div ref='child' style={{visibility: 'hidden', position: 'absolute'}}>{this.props.children}</div>;
    },

    componentDidMount(){
        $(this.refs.child).position({
            my: this.props.my,
            at: this.props.at,
            of: this.props.of(),
            collision: 'flip flip'
        }).css({visibility: 'visible'});
    }
});
```

Then use it like:
```
return    <Position my='left top' at='left bottom' of={() => this.refs.head}>
                 ...
          </Position>
```

bluetidepro:
@vladimir-rovensky hmmmm, I don't think I follow the full context of what/how you did that. Do you have a larger snippet? :/ ha Sorry for the hassle!

bluetidepro:
@vladimir-rovensky Does that also support a way to do like an "auto" position?

vladimir-rovensky:
[...]

PR #175, opened 2017-05-11
Support cross-axis positioning

devongovett:
This adds support for some additional placement values to allow alignments along the cross-axis other than center. For example, you can set placement to `"bottom right"` to align the overlay to the bottom right instead of the bottom center of a target. The default remains center, so `"bottom"` should behave the same way as it does currently. Also refactors the `calculatePosition` util to be more generic and share more code between the different directions it operates in.

positions

Should fix most of the cases #52 is asking for.


devongovett:
ping

jquense:
Sorry for let this slip through the cracks! Looks great just haven't had time to review it. @taion you're a bit more familiar with the position code wdt?

ks-s-a:
My PR (https://github.com/react-bootstrap/react-overlays/pull/176) is not needed, if this one will be merged. Please, review it.

taion:
I took a quick stab at integrating Popper, and it was a horrible dead end because of arrow positioning. Oof.

Could we make the placement naming consistent with Popper, though? i.e. `top-start` and `top-end` instead of `-left` and `-right`... opens up a path to move in that direction in the future.

Also FWIW TWBSv4 totally changes how arrows are drawn.

devongovett:
@taion sorry I missed your comment. We could change the naming, but IMO it seems weird that the cross axis names are different than the primary ones (e.g. start vs left).

devongovett:
Just added a commit to fix positioning when the overlay has a margin as well. Should fix #135.

qrohlf:
even if the cross-axis positioning isn't ready to ship yet, I'd love to see 1533b3c7a56279f39cb6bb6bcaf4efcba9ab3448 get merged and released. Happy to spin up a PR with just those changes if needed. #135 is a blocking issue for using react-overlays in my work.

taion:
Given that people are asking, I'm okay with merging this for now. Thanks for your work here.

Could you fix the names to be consistent with what Popper uses, though? I think that might make it slightly easier to migrate in the future.

chrisjaure:
Is this blocked only on the naming?

taion:
No, we're considering using Popper instead.

PR #176, opened 2017-05-16
Add horizontal scroll support for the container position calculation

ks-s-a:
Fix container position calculating and recalculating according to horizontal scroll position.

Reason of the pull request:
If we have a horizontal scroll of the container, calculations don't work properly. This fix will resolve this problem.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,3 @@
-lib/
-
 # Logs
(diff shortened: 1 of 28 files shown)


taion:
can you add a test (even a dumb one with mocks) for this?

PR #181, opened 2017-06-10
Use Popper for position calculations

### LINKED ISSUE #52: More placement types for Position

### LINKED ISSUE #135: margin on Position target causes incorrect positioning

### LINKED ISSUE #178: Auto change placement of Position when overlapped by the browser window
Hi guys,

We have a bug that tooltip overlapped by browser window. See screenshot.
Can it be possible to change "calculatePosition" method (in calculatePosition.js) to return also "placement" value. And if "overlayNode" doesn't fit in the browser windows then change "placement " to opposite value (left <-> right or top <-> bottom).

screen shot 2017-06-08 at 09 49 40

Best Regards,
Dmytro

taion:
There's a functional change here in same-axis positioning when the positioned element is about to escape. Otherwise I think everything mostly works the same. There's a debounce on the initial update, but that's not really noticeable.

I changed the API a bit to make it maybe cleaner. Plus now I explicitly hide the positioned element when we don't have a position for it.

We're not really exposing all that Popper is offering, but I'm okay with that. This closes a lot of the positioning issues on our end.

Closes #52
Closes #135
Closes #175 
Closes #176 
Closes #178


test/PositionSpec.js
```diff
@@ -1,55 +1,239 @@
...
+        await wait();
+
+        [instance.overlay1, instance.overlay2].forEach(({ props }) => {
+          expect(props.position).to.eql(expected.position);
+          expect(props.arrowPosition).to.eql(expected.arrowPosition);
+        });
+      });
+    }
+
+    [
```
taion: these are a bit different now in the "offsetBefore" and "offsetAfter" cases

test/PositionSpec.js
```diff
@@ -1,55 +1,239 @@
...
+        },
+      },
+      {
+        placement: 'left',
+        noOffset: {
+          position: { left: 50, top: 200 },
+          arrowPosition: { top: 100 },
+        },
+        offsetBefore: {
+          position: { left: 50, top: 0 },
```
taion: new

test/PositionSpec.js
```diff
@@ -59,243 +243,102 @@ describe('Position', function () {
...
-
-        });
-      });
-    }
-
-    [
-      {
-        placement: 'left',
-        noOffset: [50, 200, undefined, '50%'],
-        offsetBefore: [-200, 50, undefined, '0%'],
```
taion: corresponding old...

though note that the target position is a bit different now

the old impl would only apply the padding to cross-axis positioning. this on the other hand applies the padding to same-axis, mostly

taion:
Hmm, this makes the R-B docs page look weird even with fixes. Meh.

jquense:
Browser support? Also what is the motivation for this? I'm a big fan tho I thought we were holding off b/c of various IEs

taion:
Looks like IE10+: https://saucelabs.com/beta/builds/07fe610fcdca470f8ca45384a15409ff

No particular motivation... just saw we had a lot of position issues/PRs open, and wondered if we could scrap all of our own code. It was harder than expected and I got pretty carried away, but that happens.

Unfortunately this might not be viable – Popper takes into consideration e.g. margins on the popped elements, while TWBS3 popovers (and maybe tooltips as well?) are explicitly built with the assumption that such margins are ignored.

Masalovych:
Hi,
When we can expect this pull-request will be merged?

jquense:
@taion is this designed to account for TWB v3 reqs? I'd say we add this but ignore the v3 stuff and keep both the new and old Position around for v3 legacy but retire it here. v4 bootstrap needs the popper version anyway

taion:
This won't work with TWBSv3 without horrible hacks because of the way TWBSv3 handles margins and such.

You proposed moving the TWBSv3-compatible version of this into React-Bootstrap, which sounded like a good idea at the time.

jquense:
I'm hitting a point just in our own code where I jsut want to use popper and not the default bootstrap stuff, so i'm a bit anxious to move on that.

I think we just freeze the position component where it is and either move it to RB or keep it here as PositionLegacy, and move forward with the unhacky popper implementation

taion:
I'd move it to R-B. It's really very TWBSv3-specific, and I don't think it works too well for other cases. Or anyway the Popper-based version is more generic.

taion:
It's worth considering whether we want to use react-popper, though: https://github.com/souporserious/react-popper

The API is pretty different from what's here, but... well, it's there.

jquense:
That works for me, I'll move it over. Does this need any unhacking, or you think its good to go as a general implementation over Popper?

taion:
Config-wise, I tried to get something that pragmatically matched how TWBSv3 worked, with things like: https://github.com/react-bootstrap/react-overlays/pull/181/files#diff-b7dd97a1efefd56a539d6f75fcd04a7bR209

If we're not worrying about TWBSv3-compat any more, then I'd switch to either using better defaults, or to exposing those config flags.

bcullman:
Any chance we would be getting edge detection for free by using popper/react-popper here?

taion:
Yes.

jeffredodd:
Any progress on this? I'd love to help out but i'm not quite sure where this PR stands?

taion:
Next step is to move the legacy implementation into React-Bootstrap, because this doesn't work well with TWBSv3.

devongovett:
IMO it would be nice not to have to bring in another large dependency just for positioning. Is it really needed? What do we get that we wouldn't get from e.g. #175?

taion:
Well, it's 6 kB minified + gzipped, so not that big? Also just more powerful/flexible.

devongovett:
6K is pretty big for what it is doing I think. What actual features does it provide that you're exposing in react-overlays other than fixing cross-axis positioning?

jquense:
we are thinking about exposing all the features, thats sort of the point and benefit. :) In any case this is likely to happen for bootstrap v4 because upstream bootstrap uses it

devongovett:
@jquense is the plan to merge this soon then? I need some sort of cross-axis positioning, whether that be #175 or this.

jquense:
I'd love to merge it asap, but I don't have time at the moment to adjust the PR here per the above convo and I'm guessing @taion may be in a similar boat

devongovett:
Also is this backwards compatible with the existing Position component? If not, what is the migration path? Major version bump? We could merge #175 in the mean time which is backwards compatible...

jquense:
It doesn't need to be backwards compatible here. But we need to move the existing Position component to react-bootstrap per the convo above

devongovett:
How will that work since `Overlay` here uses `Position` here as well? We'd need to copy `Overlay` as well then, no? Or maybe make a prop to pass in the component to use... that could be useful for other custom positioning cases anyway I suppose.

adilapapaya:
@taion and @jquense would it be possible to merge #175 for the time being and save the popper integration for a larger version bump? I'm in the same boat as @devongovett – need some cross-axis positioning functionality.

p.s. Thanks for the awesome library + work so far!!

jquense:
@taion yo can we just merge this?

taion:
Well, no, because package-lock.json has a merge conflict :p

taion:
Oops finger slipped. I can look for some time to work on this though.

kumarharsh:
> Oops finger slipped. I can look for some time to work on this though.
Thank god for the finger slip! I thought for a moment this was dead 😛 
Would love it if react-overlays does support Popper - right now I'm using react-overlays as well as popper, both of them doing similar work, just that popper provides better positioning, and react-overlays provides all those other things (RootCloseWrapper, Portal, etc).

jquense:
@taion imma merge this later today

taion:
Sounds good, do you need anything from me?

jquense:
It'd be excellent if you had a moment to see if this properly sets us up for the v4 dropdowns.

PR #184, opened 2017-06-27
remove transition

jquense:


jquense:
do you know why travis is failing?

taion:
I think you need to update `package-lock.json`.

taion:
That works too I guess.

jquense:
:P the lerna thing means I don't have npm@5 installed

Issue #194, opened 2017-08-17
Modal uses non-standard keyUp listner


jeffredodd:
Currently the modal escape key listener is being fired on `keyup` which can be problematic as most key press listeners are listening for `keydown`. 

I discovered this issue when embedding a DropdownButton inside of a modal. When using keyboard navigation inside the dropdown button, attempting to close the menu with escape will also close the modal itself which creates a jarring user experience. 

It appears Twitter Bootstrap itself moved away from `keyup` to `keydown` due to similar issues. https://github.com/twbs/bootstrap/issues/13929 

I'll get a PR out in a few minutes which attempts to solve this issue and reduce a potential breaking change.

jeffredodd:
Closing issue as its been resolved by the merge of #195

PR #195, opened 2017-08-17
Change Modal esc listener to keydown from keyup

jeffredodd:
This PR changes the non-standard functionality of the esc key listener in the modal from `keyup` to `keydown` and fixes issue #194. 

The main issue can be reproduced by embedding a DropdownButton inside of a modal. When using keyboard navigation inside the dropdown button, attempting to close the menu with escape will also close the modal itself which creates a jarring user experience.


src/Modal.js
```diff
@@ -94,6 +94,12 @@ class Modal extends React.Component {
     /**
      * A callback fired when the escape key, if specified in `keyboard`, is pressed.
      */
+    onEscapeKeyDown: PropTypes.func,
+
+    /**
+     * Support for this function will be depricated. Please use `onEscapeKeyDown` instead
```
jquense: "deprecated" also can you add the `@deprecated` doclet in the comment (under the description)

src/Modal.js
```diff
@@ -428,11 +434,14 @@ class Modal extends React.Component {
     }
   }
 
-  handleDocumentKeyUp = (e) => {
+  handleDocumentKeyDown = (e) => {
     if (this.props.keyboard && e.keyCode === 27 && this.isTopModal()) {
-      if (this.props.onEscapeKeyUp) {
-        this.props.onEscapeKeyUp(e);
+      let escapeFunc = this.props.onEscapeKeyDown || this.props.onEscapeKeyUp;
```
jquense: this is a bit misleading. I'd make sure the old handler actually gets called on keyUp

src/Modal.js
jeffredodd: let me know if my recent change is the correct way to declare deprecated

jquense:
lgtm

src/Modal.js
```diff
@@ -94,6 +94,13 @@ class Modal extends React.Component {
...
      * A callback fired when the escape key, if specified in `keyboard`, is pressed.
      */
+    onEscapeKeyDown: PropTypes.func,
+
+    /**
+     * Support for this function will be deprecated. Please use `onEscapeKeyDown` instead
+     * A callback fired when the escape key, if specified in `keyboard`, is pressed.
+     * @deprecated
+     */
     onEscapeKeyUp: PropTypes.func,
```
taion: use https://github.com/react-bootstrap/prop-types-extra#deprecatedvalidator-reason maybe? that way people will get the warning; makes it easier to remove for next version

we're already pulling in `prop-types-extra`

src/Modal.js
```diff
@@ -428,12 +440,21 @@ class Modal extends React.Component {
     }
   }
 
+  handleDocumentKeyDown = (e) => {
+    if (this.props.keyboard && e.keyCode === 27 && this.isTopModal()) {
```
taion: can we do `e.key === 'Escape'`? (and below too)

i know this is an old issue, but it'd clean things up a tiny bit

taion:
Thanks!

PR #198, opened 2017-08-25
fix tooltip/popover stlying props being ignored

AlahmadiQ8:
According to the [documentation](https://react-bootstrap.github.io/components.html#tooltips-props-tooltip), I should be able to provide the following props: arrowOffsetLeft, arrowOffsetTop, positionLeft, positionTop

Unfortunately, these props are being completely ignored by the Position Element. This prevented me from for example, make the arrow 80% from the left instead of 50%. 

This pr should address the issue.


taion:
dupe https://github.com/react-bootstrap/react-bootstrap/issues/1640

Issue #188, opened 2017-07-17
Modal does not work with React 16


jochenberger:
Probably the same issue as https://github.com/facebook/react/issues/9808.
When the Modal is shown and tries to get focus (via  `componentDidUpdate` -> `onShow` - `focus`), `setModalNode` will not have been called and `getDialogElement` will return `undefined`.

jochenberger:
https://github.com/callemall/material-ui/pull/7218/files might be helpful here.

billschaller:
@jquense @jochenberger Current shipping version of react-bootstrap is still referencing react-overlay@0.7.0. That version is still using string refs, and that's not working in the Modal component for some reason with React 16.

Would you consider hotfixing this? Kind of a blocker :(

jquense:
we can't hotfix RB to v0.8 because there are a few breaking changes, we need to do it on a major bump of RB

billschaller:
Could you push an @next version to npm?

jquense:
I'm not sure that would even help no version of RO uses the new portal api yet.

string refs _should_ work still with the old API I would open an issue on the React repo since its a v16 bug

billschaller:
Ah - Part of my app was on 30.10, and the string ref was still there.

The latest release is still broken in 16 though.

It looks like the child component of Portal passed in by Modal is not mounted by the time Modal.componentDidUpdate is called - this makes sense, as AFAIK that lifecycle callback doesn't fire until the commit phase. The proper thing to do is to have the child you pass to the Portal component kick off onShow instead of doing it from cDU or cDM in the Modal component itself.

billschaller:
@jquense Sent you a chat request

jquense:
@billschaller sorry just saw the request. You are better off pinging me in the react-bootstrap discord channel https://discord.gg/0ZcbPKXt5bXLs9XK i'm `monasticpanic`

KagamiChan:
Is there any news on the landing of RO on RB? I'd like to help but I don't know what are the tasks to be done.

jochenberger:
https://github.com/facebook/react/pull/10675 is probably also relevant.

jquense:
@KagamiChan please feel free to jump in here. The tasks are to upgrade to the new createPortal api while (if possible) also supporting v15. So some amount of feature detecting and branching is needed. Portal should be the only component that needs an upgrade.

KagamiChan:
thank you for your reply, will take some time to learn about this

[...]

PR #201, opened 2017-09-15
switch to ReactDOM.createPortal

jochenberger:
Modal stops working in React 16 (#188). I thought this might fix it but it doesn't. It might be a good idea to do the switch anyway.


jochenberger:
Just switching to `ReactDOM.createPortal` doen't help, it produces the same error.

src/Portal.js
```diff
@@ -11,6 +11,9 @@ import ownerDocument from './utils/ownerDocument';
  * You can think of it as a declarative `appendChild()`, or jQuery's `$.fn.appendTo()`.
  * The children of `<Portal/>` component will be appended to the `container` specified.
  */
+
+const useCreatePortal = ReactDOM.createPortal !== undefined;
```
jochenberger: Should we also support the deprecated `ReactDOM.unstable_ createPortal`? IIRC, it's only present in React 16 pre-releases.

jochenberger:
This does not fix (#188) but might be nice to have anyway.

jquense:
Thanks KS for the help, you might want to looks at the tests in Rea t for how the new API works. It's not the same as the old one you actually call it during render()

jochenberger:
I ran the tests with React 15 and 16 and they passed. But I can have another look.

jquense:
I mean look at React's tests for createPortal,  not ours, for proper usage of the api :)

jochenberger:
Getting closer. The tests that require `React.Children.only` to throw stop working, but since that also affects `PositionSpec`, I think it's an unrelated issue.
There are still two other tests failing in `PortalSpec` though.

jochenberger:
I'm struggling with when to call `_mountOverlayTarget`. `_mountOverlayTarget` needs to find the DOM node for the container. It uses `ReactDOM.findDOMNode` via `utils/getContainer`, which means that it must not be called during render.

src/Portal.js
taion: i'd go with "no"; just support the releases i think

jochenberger:
`cDM` is too late because the overlay target needs to be initialized when we call `createPortal` from `render`.

src/Portal.js
```diff
@@ -26,9 +29,21 @@ class Portal extends React.Component {
     ])
   };
 
+  componentWillMount(){
+    if (useCreatePortal){
```
jquense: this also needs a canUseDOM guard here, for serverside rendering. dom-helpers exports that constant for eas of use

src/Portal.js
```diff
@@ -93,7 +113,12 @@ class Portal extends React.Component {
   }
 
   render() {
-    return null;
+    if (useCreatePortal && this.props.children) {
```
jquense: paired with my comment above, this needs to confirm `this._overlayTarget` exists for the SSR case

test/ModalSpec.js
```diff
@@ -171,7 +171,7 @@ describe('Modal', function () {
       backdrop.style.borderWidth).to.equal('3px');
   });
 
-  it('Should throw with multiple children', function () {
+  xit('Should throw with multiple children', function () {
```
jquense: why are these failing now?

jochenberger: I have no idea. But the failure is not caused by my changes (`PositionSpec` fails too). I suspect, there are some changes around how errors are handled and/or reported in React 16.

taion:
Given that this needs to mount its own extra portal element, I don't think you can use the cWM-based flow.

I think you need to create the element in cDM, then e.g. check `this.state` in `render` to see if we have a portal element.

jquense:
So the new Portal API is awkward. We want to render the portal element on the first render _if possible_ to avoid missing updates or bubbling issues that the older api has. But yeah, it doesn't work if the mount container is isn't just a DOM node, or in the SSR case. I'm not sure what the best option forward is...I'm leaning towards just requiring a DOM node as a container vs a component instance but that might dramatically reduce the usefulness of the component?

Side note, the behavior between the old and new API is different, createPortal, allows events to bubble through as if it was a real child

taion:
Wait why do we need `_overlayTarget` anyway here? There's no way we can safely `createElement` a `div` in cWM, is there?

jquense:
You can safely create one if you check for `document` in a browser it's going to be there, its only in SSR env that you can't but that's gonna give the same result as if you double rendered and created in cDM

src/Portal.js
jochenberger: done

src/Portal.js
jochenberger: done

test/ModalSpec.js
jochenberger: I opened #207 to track that down.

TryingToImprove:
Note that the examples does not work anymore because of `@monastic.panic/component-playground/Playground` which is still using `React.PropTypes`

TryingToImprove:
The `.to.throw` is not working because the ` assertion-error` library is not included.

jquense:
I updated `component-playground` to fix the propType issues. What do we think about this? Have we solved the "container that is a component instance" issue?

jquense:
Ok, I'm realizing there is a bunch of code here (originally) that I don't really understand any more...

jochenberger:
This is probably obsolete now that #208 is merged.

PR #202, opened 2017-09-15
WIP: Fix Portal with React 16

jochenberger:
see #188


jochenberger:
No idea why the assertion does not work anymore. It might have something to do with the new error boundaries.

PR #203, opened 2017-09-22
upgrade some dependencies

jochenberger:


package.json
```diff
@@ -36,58 +36,58 @@
...
     "react": "^0.14.9 || >=15.3.0",
     "react-dom": "^0.14.9 || >=15.3.0"
   },
   "devDependencies": {
-    "@monastic.panic/component-playground": "^2.0.0",
-    "babel-cli": "^6.24.1",
-    "babel-core": "^6.25.0",
-    "babel-eslint": "^7.2.3",
-    "babel-loader": "^7.0.0",
+    "@monastic.panic/component-playground": "^3.1.0",
```
taion: is this safe? i recall things breaking when i just did a straight upgrade here cc @jquense

package.json
```diff
@@ -36,58 +36,58 @@
...
+    "babel-cli": "^6.26.0",
+    "babel-core": "^6.26.0",
+    "babel-eslint": "^8.0.0",
```
taion: not sure about using 8.0.0 here – that one relies on babel 7, which we're not otherwise using

package.json
```diff
@@ -36,58 +36,58 @@
...
+    "react": "^15.6.1",
+    "react-bootstrap": "^0.31.3",
+    "react-dom": "^15.6.1",
     "release-script": "^1.0.2",
-    "rimraf": "^2.6.1",
+    "rimraf": "^2.6.2",
     "simulant": "^0.2.2",
-    "sinon": "^2.3.2",
-    "sinon-chai": "^2.10.0",
+    "sinon": "^3.3.0",
```
taion: karma-sinon-chai requires sinon v2.x

package.json
jochenberger: Right, the examples don't show anymore. I'll revert the upgrade.

package.json
jochenberger: I'll change it back then.

PR #204, opened 2017-09-22
Backport modal listener key change

taion:
cc @jeffredodd

### DIFF
--- a/src/Modal.js
+++ b/src/Modal.js
@@ -5,6 +5,7 @@ import contains from 'dom-helpers/query/contains';
 import canUseDom from 'dom-helpers/util/inDOM';
 import PropTypes from 'prop-types';
 import componentOrElement from 'prop-types-extra/lib/componentOrElement';
+import deprecated from 'prop-types-extra/lib/deprecated';
 import elementType from 'prop-types-extra/lib/elementType';
 import React, { cloneElement } from 'react';
 import warning from 'warning';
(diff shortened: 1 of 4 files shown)


taion:
@jquense Quick review on this? I didn't realize React-Bootstrap wasn't using v0.8.x, and this will allow people to actually see the modal a11y fix, without making us release `next`.

src/Modal.js
```diff
@@ -440,12 +456,21 @@ class Modal extends React.Component {
...
+        this.props.onEscapeKeyDown(e);
+      }
+
+      this.props.onHide();
+    }
+  }
+
   handleDocumentKeyUp = (e) => {
-    if (this.props.keyboard && e.keyCode === 27 && this.isTopModal()) {
+    if (this.props.keyboard && e.key === 'Escape' && this.isTopModal()) {
```
jeffredodd: This is a small question I could probably google and get the answer but maybe someone already knows. Will this key always be `'escape'` even on a Chinese keyboard for example? Just concerned about localization when hard coding english words in the codebase.

taion: Yup, see https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/key/Key_Values

(Note `Escape`, never `escape` 😛)

taion:
Released in v0.7.1.

taion:
(so just reinstall deps for React-Bootstrap as needed and you should be good)

PR #206, opened 2017-09-25
Target margin handling

### LINKED ISSUE #135: margin on Position target causes incorrect positioning

qrohlf:
Correctly handle the case where a target element has margins, with options to position the overlay either inside or outside the margins of the target.

Resolves #135

### DIFF
--- a/src/Overlay.js
+++ b/src/Overlay.js
@@ -33,7 +33,6 @@ class Overlay extends React.Component {
       , target
       , placement
       , insideMargin
-      , legacy
       , shouldUpdatePosition
       , rootClose
       , children
@@ -53,7 +52,7 @@ class Overlay extends React.Component {
     // Position is be inner-most because it adds inline styles into the child,
     // which the other wrappers don't forward correctly.
     child = (
-      <Position {...{container, containerPadding, target, placement, insideMargin, legacy, shouldUpdatePosition}}>
+      <Position {...{container, containerPadding, target, placement, insideMargin, shouldUpdatePosition}}>
         {child}
       </Position>
     );
--- a/src/Overlay.js
+++ b/src/Overlay.js
@@ -32,6 +32,8 @@ class Overlay extends React.Component {
       , containerPadding
       , target
       , placement
+      , insideMargin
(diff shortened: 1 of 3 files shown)


taion:
Are you using the non-`insideMargin` code path? Looks like TWBS just always uses the equivalent of your `insideMargin = true`. If that's good enough for you, then I'd prefer to not expose the configuration knob and always just position the overlay inside the margin.

qrohlf:
@taion I'm not using the `insideMargin` code path. Based on the diff if I omit the `insideMargin` logic, it looks like another PR may have already fixed the 'legacy behavior' shown above and rendered this PR unnecessary. I'll drop `master` into my app and verify when I get the chance, but I suspect we can close this PR.

taion:
I'm not aware that we've merged anything here, so it'd be weird if the behavior changed.

Actually, that's a little confusing. Your OP shows a change in behavior for `insideMargin = false`, but I agree that I don't see any code that does that. How does that work?

Sorry for taking so long to get this this.

qrohlf:
@taion OP was rendered with 1fac36a, and there was an issue with that revision where I confused the tooltip offsetParent margins with the target's margins, since they happened to match. 

This whole PR is probably the result of a PEBKAC error on my part, which I'll verify today or tomorrow. If that's the case, I'll close it and owe you a beer for confusing all parties involved ;)

PR #207, opened 2017-09-26
WIP: Test failures with React 16

jochenberger:
Some tests start to fail with React 16.

### DIFF
--- a/test/ModalSpec.js
+++ b/test/ModalSpec.js
@@ -11,6 +11,16 @@ import { render, shouldWarn } from './helpers';
 
 const $ = componentOrNode => jQuery(ReactDOM.findDOMNode(componentOrNode));
 
+class ErrorBoundary extends React.Component {
+  // React >= 16
+  componentDidCatch(error, info) {
+    this.props.onError(error, info);
+  }
+
+  render() {
+    return this.props.children;
+  }
+}
 
 describe('Modal', function () {
   let mountPoint;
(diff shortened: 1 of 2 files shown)


jochenberger:
I'll disable the Modal tests, we're addressing that issue in #201.

jochenberger:
I stole the error boundary part from #208. Unfortunately, it doesn't work with React 15 yet. I guess the `unstable_handleError` does not prevent the error from being thrown. Is this even relevant or will you upgrade to 16 anyway?

jquense:
We can just update the dev dep to 16 for this I think. Does the error boundary approach work?

taion:
We'd need to add an assertion on the number of `expect` calls triggered.

That or save off the exception and do the checking outside `render` (does that work given that `render` is potentially async?).

Otherwise those test cases would pass even if the error is not thrown.

jochenberger:
@jquense, unfortunately, it doesn't seem as if the approach works.

jquense:
some more context https://github.com/facebook/react/issues/10474#issuecomment-332823888

taion:
If we're not testing lifecycle handling here, can we just use the SSR API? That still throws normally, right?

jochenberger:
I can't get it to work with React 16. It seems as if `componentDidCatch` is never called. The behavior with `react-dom/server` is basically the same.

taion:
Doesn't it explicitly throw a normal JS exception with `react-dom/server`?

jochenberger:
Closing in favor of #216

PR #208, opened 2017-09-27
Update Portal

jquense:
Our tests are SO FLAKEY.

@jochenberger sorry i was working off your PR but needed to see what was required with just the new API. I think we can still switch between LegacyPortal and Portal for back compat.

I don't love how this turned out I had to add a callback to Portal because it takes two render cycles to get something rendered in order to fire the focus() callback


jochenberger:
Yeah, no worries, I just wanted to make some progress on the issue.

package.json
```diff
@@ -44,7 +44,7 @@
     "react-dom": "^0.14.9 || >=15.3.0"
   },
   "devDependencies": {
-    "@monastic.panic/component-playground": "^2.0.0",
+    "@monastic.panic/component-playground": "^3.2.2",
```
jochenberger: Wouldn't it be better to do that upgrade in a separate PR? Would be easier to review IMO.

test/ModalSpec.js
```diff
@@ -394,15 +411,15 @@ describe('Modal', function () {
...
     });
 
     it('Should return focus to the modal', () => {
-      document.activeElement.should.equal(focusableContainer);
+      expect(document.activeElement).to.equal(focusableContainer);
 
       render(
         <Modal show>
-          <div className='modal'>
+          <div className='modal'>q
```
jochenberger: using `vi`? ;-)

test/ModalSpec.js
```diff
@@ -11,6 +11,16 @@ import { render, shouldWarn } from './helpers';
 
 const $ = componentOrNode => jQuery(ReactDOM.findDOMNode(componentOrNode));
 
+
+class ErrorBoundary extends React.Component {
```
jochenberger: Is this related to the `React.Children.only` test failures (#207)? If so, a separate PR might be a good idea.

test/PositionSpec.js
```diff
@@ -20,7 +20,7 @@ describe('Position', function () {
     assert.equal(ReactDOM.findDOMNode(instance).nodeName, 'SPAN');
   });
 
-  it('Should warn about several children', function () {
+  xit('Should warn about several children', function () {
```
jochenberger: Shouldn't that be `it` again?

test/ModalSpec.js
```diff
@@ -171,15 +184,20 @@ describe('Modal', function () {
       backdrop.style.borderWidth).to.equal('3px');
   });
 
-  it('Should throw with multiple children', function () {
-    expect(function(){
-      render(
+  xit('Should throw with multiple children', function () {
```
jochenberger: Shouldn't that be `it` again?

test/ModalSpec.js
```diff
@@ -353,8 +371,7 @@ describe('Modal', function () {
     });
 
     it('Should focus on the Modal when it is opened', function () {
-
-      document.activeElement.should.equal(focusableContainer);
+      expect(document.activeElement).to.equal(focusableContainer);
```
jochenberger: Does that make a difference?

test/ModalSpec.js
```diff
@@ -11,6 +11,16 @@ import { render, shouldWarn } from './helpers';
 
 const $ = componentOrNode => jQuery(ReactDOM.findDOMNode(componentOrNode));
 
+
+class ErrorBoundary extends React.Component {
+  componentDidCatch(error, info) {
```
jochenberger: Do you want to upgrade the React dev-dependencies to 16? Otherwise you want want to add support for `unstable_handleError`.

test/ModalSpec.js
jquense: it seems to locally. I'm not at all sure why

package.json
jquense: I updated this to verify stuff was working apart from the tests which are flaky for me. I think it should be ok

src/Modal.js
```diff
@@ -472,7 +475,7 @@ class Modal extends React.Component {
     let autoFocus = this.props.autoFocus;
     let modalContent = this.getDialogElement();
     let current = activeElement(ownerDocument(this));
-    let focusInModal = current && contains(modalContent, current);
+    let focusInModal = modalContent && current && contains(modalContent, current);
```
taion: see https://github.com/react-bootstrap/react-bootstrap/issues/2726#issuecomment-329047753

i think especially w/react 16 this might be a little flaky, since `modalContent` might not be the actual modal content... or at least not all of it

src/Portal.js
```diff
@@ -111,7 +70,6 @@ class Portal extends React.Component {
 
     return null;
   }
-
 }
 
-export default Portal;
+export default ReactDOM.createPortal ? Portal : LegacyPortal;
```
taion: neat

src/Portal.js
```diff
@@ -5,7 +5,7 @@ import ReactDOM from 'react-dom';
 
 import getContainer from './utils/getContainer';
 import ownerDocument from './utils/ownerDocument';
-
+import LegacyPortal from './LegacyPortal'
```
taion: newline after this

src/Portal.js
```diff
@@ -23,81 +23,40 @@ class Portal extends React.Component {
...
-    }
-  }
-
-  _unrenderOverlay = () => {
-    if (this._overlayTarget) {
-      ReactDOM.unmountComponentAtNode(this._overlayTarget);
-      this._overlayInstance = null;
-    }
+  setContainer = (props = this.props) => {
+    this._portalContainerNode = getContainer(props.container, ownerDocument(this).body);
```
taion: why not just use `setState` here?

taion: can we handle setting this in the ctor in cases where it is defined there (for whatever reason?)

src/LegacyPortal.js
```diff
@@ -0,0 +1,124 @@
...
+    let overlay = !this.props.children
+      ? null
+      : React.Children.only(this.props.children);
+
+    // Save reference for future access.
+    if (overlay !== null) {
+      this._mountOverlayTarget();
+      ReactDOM.unstable_renderSubtreeIntoContainer(
+        this, overlay, this._overlayTarget, (inst) => {
+          this._overlayInstance = inst
```
taion: semi

src/LegacyPortal.js
```diff
@@ -0,0 +1,124 @@
...
+
+          if (this.props.onRendered) {
+            this.props.onRendered()
```
taion: semi

src/Portal.js
taion: hmm, is there any meaningful impact to no longer creating a `<div>` here? what if the container does end up being `document`?

test/PortalSpec.js
```diff
@@ -76,29 +70,8 @@ describe('Portal', function () {
     assert.equal(instance.overlay.getOverlayDOMNode(), null);
   });
 
-  it('Should render only an overlay', function() {
```
taion: why does this test case no longer work?

examples/App.js
```diff
@@ -65,8 +65,8 @@ class ExampleEditor extends React.Component {
         lang="js"
         theme="neo"
         scope={scope}
-        codeText={this.props.codeText}
-        collapsableCode
+        code={this.props.codeText}
+        collapsable
```
taion: eww it's not "collapsible" like in react-bootstrap? ;p

test/ModalSpec.js
taion: could be flakiness with the prototype manipulation stuff

src/Modal.js
jquense: i'm not sure I understand that, don't we control this DOM node?

src/Portal.js
jquense: I don't think i understood what the div was doing in the old code, the container exists somewhere why were we then appending it to a different parent?

test/PortalSpec.js
jquense: i forget...

src/Portal.js
jquense: I'd like an earlier path but it takes some reworking. right now we do `findDOMNode` on the result of the function or the value, and we can't reasonable know when to not do that, or when it will error. I think we need to change the API to provide a DOM node specifically vs a component instance. e.g. `() => findDOMNode(this)`, i think we do that for `target` in Position so it's probably a good idea anyway

taion: Maybe just fast-path `container && container.nodeType === Node.ELEMENT_NODE`?

src/Modal.js
taion: No – it's supposed to be the DOM node corresponding to `children`, but the actual node could be e.g. a comment or something per the linked React-Bootstrap issue if e.g. `children` don't end up rendering anything. Perhaps it could be something even weirder with React 16 – like a text node or something.

src/Portal.js
taion: I think the `<div>` is like – if the container ends up being `document`, you don't want to try to mount something on `document`.

jquense: @taion there are too many issues with doing a fast path. For instance, we default `container` to the owner `document`, which we get by doing `findDOMNode` on the component instance (Portal/Modal), in the early case that won't work because the component hasn't mounted yet.

these can can be individually handled, but gets harder across components, e.g Portal mounts early, and calls back to Modal's `onShow` which assumes/needs the owner document

taion: Okay, forget the fast path then.

What are your thoughts around adding the `<div>`? Is that just a legacy thing to prevent trying to e.g. unmount the tree rooted at `document`?

jquense: I'm not sure honestly but it seems safe to remove. if someone wants to try and mount over their application that's sort of their problem :P my guess as to why it's there is so you don't blow away the dom that's already there when you render() but createPortal already appends into it's container

taion: okay, so just to confirm, nothing will go wrong with using `document` as the portal node?

test/PositionSpec.js
taion: ping?

jquense: this is failing because of the other issues trying to be fixed in #207

src/Modal.js
jquense: i'm not sure how to deal with this without supplying our own wrapping div, children should render a single element

src/Portal.js
jquense: having the container be `document` wouldn't work before either. if it's document.body or any thing else i'll work fine

jquense:
lets get this in 👍

jharris4:
I tried this out (with React 16 and react-bootstrap next branch) in a project I have that's using modals, and it's looking good to me.

Let me know if there's anything I can do to help test this out

src/Modal.js
taion: we'd make a `<RefHolder>` component, attach a `ref` to it, then `findDOMNode` it... but this isn't really a regression or anything, so i guess this is fine for now

test/PositionSpec.js
taion: add a FIXME

PR #211, opened 2017-10-06
Use keyCode instead of key

### LINKED ISSUE #210: Not working, close the modal on press escape key in IE11
I doubt this commits  67d06fd92b7ecb29477fd39eea8767ba16242762.
Because KeyboardEvent.key is 'Esc', not 'Escape' in IE11, so you need to look at KeyboardEvent.keyCode.

rbesc

taion:
Fixes #210

Sorry @jeffredodd!

### DIFF
--- a/src/Modal.js
+++ b/src/Modal.js
@@ -445,7 +445,7 @@ class Modal extends React.Component {
   }
 
   handleDocumentKeyDown = (e) => {
-    if (this.props.keyboard && e.key === 'Escape' && this.isTopModal()) {
+    if (this.props.keyboard && e.keyCode === 27 && this.isTopModal()) {
       if (this.props.onEscapeKeyDown) {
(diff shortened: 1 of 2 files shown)


jquense:
I don't think this fixes it...React normalizes "Esc" -> "Escape" https://github.com/facebook/react/blob/master/src/renderers/dom/shared/utils/getEventKey.js

taion:
It does, but this is the document keydown event, so it's not going through React.

jquense:
ooo

jeffredodd:
Haha @taion I knew something was weird about using `key`! Now we all know! 😄

taion:
It would be the correct thing if it were a React event, I think, since React normalizes `key`, plus `keyCode` is deprecated, but I think this is the best option for handler a browser event for now.

PR #215, opened 2017-10-18
Modal container position should be fixed

jmera-cc:
This CSS prevents the modal container from scrolling

### Steps to reproduce
1. Navigate to https://react-bootstrap.github.io/react-overlays/#modals on a mobile device
2. Open a modal
3. You should be able to scroll up/down

### DIFF
--- a/src/ModalManager.js
+++ b/src/ModalManager.js
@@ -23,12 +23,13 @@ function findContainer(data, modal) {
 }
 
 function setContainerStyle(state, container) {
-  let style = { overflow: 'hidden' };
+  let style = { overflow: 'hidden', position: 'fixed' };
 
   // we are only interested in the actual `style` here
   // becasue we will override it
   state.style = {
     overflow: container.style.overflow,
+    position: container.style.position,
     paddingRight: container.style.paddingRight
   }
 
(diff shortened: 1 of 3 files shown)


jmera-cc:
I'm getting a test failure & can't seem to figure out what's wrong; any help with someone familiar with the codebase would be appreciated.

jmera-cc:
It looks like recent changes have fixed tests that were failing locally (but passing on CI). This branch is up-to-date, everything looks green. Would someone review?

jquense:
Hi, thanks for the PR I'm not sure I understand what this is solving. We already have props for customizing the container styles, and I don't know that this is a good default style

jmera-cc:
Modals are supposed to disable scrolling of the page content while open. I've noticed that on mobile devices this doesn't work as expected. I think this might be related to https://github.com/twbs/bootstrap/issues/7501

jmera-cc:
https://www.dropbox.com/s/1lfvllx37agnyki/react-overlays.mov?dl=0

jmera-cc:
Although this change works on mobile, `position: fixed` has a side effect of scrolling the user to the top of the page which we don't want.

Further reading:
https://github.com/twbs/bootstrap/issues/15852

PR #218, opened 2017-10-19
Only call onRendered from LegacyPortal on initial render

### LINKED ISSUE #217: Modals with input are broken with pull request #208 in React 15
Hey guys,

When my project upgraded to react-overlays 0.7.3, my modals with input forms all broke.  When I change the value of an input field, the component is rendering.  It seems that every time the component renders, you are calling a focus() on the modal.  So when I tab out of my field, instead of tabbing into the next field, the focus goes to the modal form and I lose the input focus.

As far as I can tell, this code in Modal.js is the culprit:

```
  onPortalRendered = () => {
    this.focus();

    if (this.props.onShow) {
      this.props.onShow();
    }
  }
```

I don't know if something different happens in 16, but I am using 15 and this is causing a problem.  The old code only called focus onShow.  Why did this change?

So far, my only solution is to set autoFocus=false and enforceFocus=false, but I am not sure if that is ideal.

Thanks,
Dave

taion:
Fixes #217

Untested (since our tests don't run against React 15).

@bunkscene Do you want to test this locally?

### DIFF
--- a/src/LegacyPortal.js
+++ b/src/LegacyPortal.js
@@ -76,11 +76,12 @@ class Portal extends React.Component {
     // Save reference for future access.
     if (overlay !== null) {
       this._mountOverlayTarget();
-      ReactDOM.unstable_renderSubtreeIntoContainer(
-        this, overlay, this._overlayTarget, (inst) => {
-          this._overlayInstance = inst;
 
-          if (this.props.onRendered) {
+      const initialRender = !this._overlayInstance;
+
+      this._overlayInstance = ReactDOM.unstable_renderSubtreeIntoContainer(
+        this, overlay, this._overlayTarget, () => {
+          if (initialRender && this.props.onRendered) {
             this.props.onRendered();
           }
         }


bunkscene:
Ah, thanks for the quick look. Yeah I’ll test it later tonight to tomorrow AM.

taion:
When I say "untested", I mean I literally didn't test this at all, BTW, not just that I didn't add a test case – but I think something like this is the right way to handle this case.

bunkscene:
I like your style.

bunkscene:
I tried this and it's not working for me.  The inst variable is always undefined, so the _overlayInstance is never getting set.  I am not super familiar with this method, but maybe I can do something in the modal definition to get this to work right?

jquense:
are you sure you are using react 15? that's pretty weird, it wouldn't work at all if it was undefined

bunkscene:
15.6.2.  Trying to get a sample project up to show this.

taion:
Well, we don't really use `_overlayInstance` at all, and that's what I was checking.

jquense:
its used for tests i think...

taion:
Okay this is fixed properly now. PTAL

src/LegacyPortal.js
```diff
@@ -76,11 +76,12 @@ class Portal extends React.Component {
     // Save reference for future access.
     if (overlay !== null) {
       this._mountOverlayTarget();
-      ReactDOM.unstable_renderSubtreeIntoContainer(
-        this, overlay, this._overlayTarget, (inst) => {
-          this._overlayInstance = inst;
```
taion: turns out this is the react component instance... since the immediately rendered thing for the modal is a `<div>`, it ends up being `undefined`

src/LegacyPortal.js
```diff
@@ -76,11 +76,12 @@ class Portal extends React.Component {
...
 
-          if (this.props.onRendered) {
+      const initialRender = !this._overlayInstance;
+
+      this._overlayInstance = ReactDOM.unstable_renderSubtreeIntoContainer(
```
taion: this (as before) ends up being the DOM node though

bunkscene:
Yes! LGTM.

src/LegacyPortal.js
jquense: i think i moved this coe around as a hotfix for v16 but doesn't matter now oops

PR #220, opened 2017-10-20
Use more robust method of getting dialog element

taion:
Note base branch.

Mostly fixes https://github.com/react-bootstrap/react-bootstrap/issues/2600
Ref https://github.com/react-bootstrap/react-overlays/pull/208/files#r142582319

Sorry for the random formatting changes on the tests. Got carried away.

Perhaps we should still check for `.hasAttribute` on `dialogElement`, though – in the case where it's e.g. a text node, the dialog element will just be a string and thus won't have that method.

### DIFF
--- a/test/ModalSpec.js
+++ b/test/ModalSpec.js
@@ -419,7 +419,6 @@ describe('<Modal>', () => {
         </Modal>
         , focusableContainer);
 
-
       focusableContainer.focus();
 
       document.activeElement.className.should.contain('modal');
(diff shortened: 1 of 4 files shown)


test/ModalSpec.js
```diff
@@ -434,6 +436,15 @@ describe('Modal', function () {
         </Modal>
         , focusableContainer);
     });
-  });
 
+    it('should not attempt to focus nonexistent children', () => {
+      const Dialog = () => null;
```
taion: The old logic found some weird container `<div>` (maybe from a transition or something?) as the dialog element, instead of picking up on there not being any dialog element.

PR #222, opened 2017-10-23
Drop busted getOverlayDOMNode

taion:
It actually doesn't even do anything on the non-legacy portal. Plus it's broken on the legacy portal half the time anyway. Since we already broke it, and we don't document it, I'm going to say this was some internal test-only thing, and it's not semver-breaking to just remove it entirely.


ericgio:
This broke some of my code. I understand it was an undocumented method, but it certainly looks like a public method in the code (no prepended underscore). Minor version bump would have been nice here.

taion:
It was non-functional in the previous commit anyway. Just use `findDOMNode` or attach your own ref if you need the node.

PR #228, opened 2017-11-27
Modal: add unmountOnExit for backdrop transition

lixiaoyan:
Fix a flick issue with React 16.

### DIFF
--- a/src/Modal.js
+++ b/src/Modal.js
@@ -328,6 +328,7 @@ class Modal extends React.Component {
       backdrop = (
         <Transition
           appear
+          unmountOnExit
           in={this.props.show}
         >
           {backdrop}


taion:
Can you clarify? Don't we already unmount the entire modal when it's hidden?

lixiaoyan:
@taion That's right, but React 16 is async: that means these two action may be not processed at the same time.

lixiaoyan:
For example: `<CSSTransition />`. DOM manipulation is sync but React update is async. There may be zero or one (or more!) frames between these two actions.

taion:
I understand that, but how do you get flicker? Isn't the non-shown backdrop invisible? Is there some visual artifact from not unmounting it?

lixiaoyan:
@taion After it exited, all transition class names has been removed from the dom, and it shows again.

taion:
It should be rendering in the "out" state, though. In that state, it should have opacity 0, no?

lixiaoyan:
```
// state 1
class="..."

// When entering
class="... Transition-enter"
// next frame
class="... Transition-enter Transition-enter-active"
// timeout
class="..."

// state 2
class="..."

// When exiting
class="... Transition-exit"
// next frame
class="... Transition-exit Transition-exit-active"
// timeout
class="..."

// state 3
class="..."
```

There's no different between state 1, 2 and 3, it just shows.

taion:
Okay, I think I understand what you're saying. @jquense does this mean that this example is a bit busted? https://github.com/reactjs/react-transition-group/blob/7a641148849fc3aa0aa0acb673da9bc3cc35a7a0/stories/ReplaceTransition.js#L9-L38

If I'm understanding the code correctly, this will render with no opacity property at all when it's in the exited state?

lixiaoyan:
https://reactcommunity.org/react-transition-group/#CSSTransition See this example.

taion:
Right, but that example clearly isn't doing what it's supposed to. The box isn't supposed to flicker.

lixiaoyan:
@taion I think it's behavior is just following the original implement of `<CSSTransitionGroup />` in react-addons-css-transition-group.

taion:
@jquense is the primary maintainer on RTG so he can state for sure what's intended (:

jquense:
The plan css transition api is sort of busted yeah. We need and exit-done and enter-done classes, the current API is a copy from v1 and basically only works well in the context of  Transition group.

The plain Transition component tho doesn't have this problem which is what are using and why we don't see this in RB. The "right" fix for this with custom CssTransitions is to fix RTG tho, I wouldn't do it here

taion:
So what should we do here?

jquense:
upstream a fix? There isn't really a bug here yeah?

taion:
Like add a `exit-done` thing? Or is the transition itself buggy? Like maybe that `<Fade>` example should define a "base" state with opacity 0?

jquense:
yeah an `exit-done` this shouldn't affect RB's transition, they use either the raw Transition component or our old Transition component which covered these cases

taion:
Should we close this, then?

PR #229, opened 2017-11-28
Create portal immediately when possible

### LINKED ISSUE #224: Refs on elements in Portal are not usable in parent componentDidUpdate in React 16
Ref https://github.com/react-bootstrap/react-bootstrap/issues/2841

This is a regression introduced here with the change to the React 16 portal API. With the new portal API, the contents of the portal don't actually get mounted until after the parent finishes rendering, so refs there aren't available in `componentDidUpdate` on the parent.

To resolve this, I think we actually need to do the DOM stuff around the portal when rendering the component initially in React 16.

taion:
Fixes #224

### DIFF
--- a/src/Portal.js
+++ b/src/Portal.js
@@ -1,3 +1,4 @@
+import canUseDom from 'dom-helpers/util/inDOM';
 import PropTypes from 'prop-types';
 import componentOrElement from 'prop-types-extra/lib/componentOrElement';
 import React from 'react';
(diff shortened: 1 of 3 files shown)


src/Portal.js
```diff
@@ -29,26 +30,49 @@ class Portal extends React.Component {
...
+    if (!canUseDom) {
+      return;
+    }
+
+    let { container } = this.props;
+    if (typeof container === 'function') {
+      container = container();
+    }
+
+    if (container && !ReactDOM.findDOMNode(container)) {
```
jquense: doesn't this throw if the thing isn't a component instance?

PR #237, opened 2018-02-24
Affix - Fixed the problem of page flashing when scroll page.

AlingChen:
If the Affix element is fixed, the scrollTop will be reduced, and the `positionTopMin` may be less than `offsetTop`, Affix element will back to the original position. so the page will flash when repeated scrolling.


taion:
Thanks for the PR. Do you have an example (you can use the editable examples on our docs page) showing this behavior? I don't think I've seen this happen before.

PR #251, opened 2018-06-29
Modal manager improvements

jquense:


jquense:
Yeah, see the react-bootstrap PR, v4 adjusts all fixed position elements in the container on modal show, the simplest way to implement was to extend and override

PR #252, opened 2018-07-02
Dropdown

jquense:
This is now sufficiently general that I think it makes sense here. Plus i need a generic dropdown component


package.json
```diff
@@ -36,11 +36,15 @@
     "testonly": "npm run tdd -- --single-run"
   },
   "dependencies": {
+    "@react-bootstrap/react-popper": "^1.0.2",
```
jquense: we'll use the real one when they merge my PR

src/DropdownMenu.js
```diff
@@ -0,0 +1,141 @@
+import PropTypes from 'prop-types'
+import React from 'react'
+
+import { Popper } from '@react-bootstrap/react-popper'
+import DropdownContext from './DropdownContext'
+import RootCloseWrapper from './RootCloseWrapper'
+import mapContextToProps from 'react-context-toolbox/lib/mapContextToProps'
+
+class DropdownMenu extends React.Component {
+  static displayName = 'ReactOverlaysDropdownMenu'
```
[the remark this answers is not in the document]
jquense: idk yet. I think it maybe makes sense to prefix displayNames generally so stuff is more readable in the dev tools, w/d/t?

src/DropdownMenu.js
```diff
@@ -0,0 +1,141 @@
...
+      flip,
+      menuRef,
+      alignRight,
+      drop,
+      toggleNode,
+      rootCloseEvent,
+      popperConfig = {},
+    } = this.props
+
+    let placement = alignRight ? 'bottom-end' : 'bottom-start'
```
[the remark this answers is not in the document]
jquense: yeah probably. I just took this from RB

package.json
jquense: They recently redid the API, and it's essentially what i've have written anyway, so seems like a nice win to use the upstream package, esp. since it's maintained (now) by the popper.js guy

taion: great!

PR #256, opened 2018-07-27
Overlays

jquense:


src/Dropdown.js
```diff
@@ -51,10 +52,12 @@ const propTypes = {
 }
 
 const defaultProps = {
-  as: 'div',
   itemSelector: '* > *',
 }
 
+/**
```
taion: ?

src/DropdownMenu.js
```diff
@@ -31,7 +31,7 @@ class DropdownMenu extends React.Component {
      */
     flip: PropTypes.bool,
 
-    usePopper: PropTypes.oneOf([true, false, 'lazy']),
+    usePopper: PropTypes.oneOf([true, false]),
```
taion: `PropTypes.bool` then?

src/Overlay.js
```diff
@@ -27,15 +33,36 @@ class Overlay extends React.Component {
     return null
   }
 
+  componentDidMount() {
+    this.setState({ target: this.getTarget() })
+  }
+
+  componentDidUpdate(prevProps) {
+    if (this.props === prevProps) return
```
taion: it's really ok/meaningful to do a ref equality check on all props?

shouldn't the props _always_ change?

src/Overlay.js
```diff
@@ -48,43 +75,56 @@ class Overlay extends React.Component {
...
 
     let child = children
+    const popperProps = {
+      placement,
+      referenceElement: this.state.target,
+      enableEvents: props.show,
+      modifiers: {
+        ...modifiers,
+        preventOverflow: {
+          padding: containerPadding || 5,
```
taion: this `5` should be a prop?

src/WaitForContainer.js
```diff
@@ -0,0 +1,65 @@
...
+class WaitForContainer extends React.Component {
+  constructor(...args) {
+    super(...args)
+
+    if (!canUseDom) return
+
+    let { container } = this.props
+
+    if (typeof container === 'function') container = container()
+    if (container && !ReactDOM.findDOMNode(container)) {
```
taion: O_O i'm surprised this is safe

src/Overlay.js
jquense: It's a heuristic for "just state changed"  which is the only time they don't change

src/Overlay.js
jquense: I guess yeah, it's the popper default

src/WaitForContainer.js
jquense: Yeah me too...I had it wrapped in a try/catch but no errors were ever thrown so idk

src/Overlay.js
taion: ah ok... we can't just use `undefined` then?

src/Dropdown.js
jquense: Didn't compile otherwise, I need to add docs tho

src/DropdownMenu.js
jquense: Lol

PR #259, opened 2018-07-28
old school esm

jquense:
maybe we should try this? https://developers.livechatinc.com/blog/how-to-create-javascript-libraries-in-2018-part-2#proxy-directories

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 lib/
+es/
 .cache/
 www/public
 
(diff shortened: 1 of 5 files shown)


taion:
isn't the suggested method going to break horribly with the native ESM support in node, though?

jquense:
i don't think anymore than our current `lib/`, or `es/` approach. I mean ultimately i think none of this will matter in node since it doesn't use the `module` main, only `.mjs` as the parse target indicator.

who. knows. tho. :(

PR #261, opened 2018-08-10
fix: clean up overlay a bit more

jquense:


jquense:
@taion what do you think about this for a render prop format generally, anything that is meant to be spread through is on `props`.

src/Dropdown.js
```diff
@@ -202,7 +224,7 @@ class Dropdown extends React.Component {
     return (
       <DropdownContext.Provider value={this.state.context}>
         <Popper.Manager>
-          {children({ onKeyDown: this.handleKeyDown })}
+          {children({ props: { onKeyDown: this.handleKeyDown } })}
```
taion: I think I get what you're trying to do here, but I'm a little confused by this API.

For example, does it really make sense for `onKeyDown` here to be under `props`, but for `onToggle` below to not be thus?

jquense: My thought was to try and make crystal clear what items to spread to a host element and have a place to add additional arguments withough breaking existing components that are spreading. We should change on toggle to just toggle maybe if that's clearer?

taion: it's weird that we force you into this `onKeyDown` thing without, say, doing the same for `onClick` on toggle :p

taion: but i guess renaming it from `onToggle` to `toggle` would work!

PR #272, opened 2018-10-30
Render context consumer instead of context itself

nathanstitt:
React 16.6 has a deprecation warning about rendering context directly

Fixes https://github.com/react-bootstrap/react-bootstrap/issues/3347

### DIFF
--- a/src/DropdownMenu.js
+++ b/src/DropdownMenu.js
@@ -189,7 +189,7 @@ class DropdownMenu extends React.Component {
 }
 
 const DecoratedDropdownMenu = mapContextToProps(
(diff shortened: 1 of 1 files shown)


taion:
don’t we just want to upgrade react context toolbox?

jquense:
probably, but it was easier to just accept:P

pankajdkoli:
When to Use Context that time i got erorr.
react-dom.development.js:67 Warning: Rendering <Context> directly is not supported and will be removed in a future major release. Did you mean to render <Context.Consumer> instead?
    at App

Issue #281, opened 2019-01-31
Add @babel/plugin-transform-runtime


vtaits:
https://babeljs.io/docs/en/next/babel-plugin-transform-runtime.html

You can reduce size of package using this plugin. For example:

Before

```
function _extends() { _extends = Object.assign || function (target) { for (var i = 1; i < arguments.length; i++) { var source = arguments[i]; for (var key in source) { if (Object.prototype.hasOwnProperty.call(source, key)) { target[key] = source[key]; } } } return target; }; return _extends.apply(this, arguments); }
```

After

```
import _extends from "@babel/runtime/helpers/extends";
```

bpas247:
It seems #296 closes this issue.

PR #282, opened 2019-01-31
Add @babel/plugin-transform-runtime

vtaits:
#281

### DIFF
--- a/.babelrc.js
+++ b/.babelrc.js
@@ -14,6 +14,7 @@ module.exports = api => ({
   ],
   plugins: [
     ['@babel/plugin-proposal-class-properties', { loose: true }],
+    '@babel/plugin-transform-runtime',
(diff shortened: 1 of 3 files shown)


taion:
do you have numbers on what this does for the bundle size? my impression is that the helpers compress really well, so the net benefit of runtime when we don't need its kind of polyfilling is quite low

jquense:
It should be a clearer win with Babel 7 I think since the runtime doesn't polyfill anything by default anymore. But we should check

vtaits:
I have screens before and after injecting this plugin to my project.

Before:
image

After:
image

But bundle contains also polyfills and used libraries.

If every library will import same helpers from `@babel/runtime` bundle size will more small.

vtaits:
With `react-overlays`: 280K before, 272K after.

PR #287, opened 2019-03-14
feat: Remove affix support

taion:


taion:
I also removed `getDocumentHeight` and `ownerWindow`, which were only used by the affix components.

PR #288, opened 2019-03-15
feat: Hooks and API simplifications

jquense:


src/DropdownMenu.js
```diff
@@ -109,23 +98,23 @@ class DropdownMenu extends React.Component {
...
-    } = this.props;
+    const { flip, usePopper, rootCloseEvent, popperConfig = {} } = this.props;
+
+    const { drop, menuRef, toggleNode } = this.context;
+
+    const show =
+      this.context.show == null ? this.props.show : this.context.show;
+
+    const alignEnd =
+      this.context.alignEnd == null
```
taion: shouldn't props win over context in general?

src/DropdownMenu.js
```diff
@@ -188,17 +177,13 @@ class DropdownMenu extends React.Component {
...
-    toggleNode,
-    onToggle: toggle,
-    show: show == null ? props.show : show,
-    alignEnd: alignEnd == null ? props.alignEnd : alignEnd,
-  }),
-  DropdownMenu,
-);
-
-export default DecoratedDropdownMenu;
+const f = ({ show, alignEnd, toggle, drop, menuRef, toggleNode }, props) => ({
```
taion: we should be exporting this `f`? or at least using it?

taion: or maybe we're not using this at all?

src/DropdownToggle.js
```diff
@@ -21,22 +33,15 @@ const propTypes = {
    */
   children: PropTypes.func.isRequired,
 };
+
 function DropdownToggle({ children }) {
```
taion: is this essentially soft-deprecated in favor of the hook now?

src/Overlay.js
```diff
@@ -1,166 +1,126 @@
...
+  const popperProps = {
+    ...popperConfig,
+    placement,
+    innerRef: ref,
+    referenceElement: target,
+    enableEvents: props.show,
+    modifiers: {
+      ...modifiers,
+      preventOverflow: {
+        padding: containerPadding || 5,
```
taion: can we use `defaultProps` here?

www/gatsby-config.js
```diff
@@ -21,7 +21,7 @@ module.exports = {
         name: 'source',
       },
     },
-    'gatsby-plugin-emotion',
+    // 'gatsby-plugin-emotion',
```
taion: ?

src/DropdownMenu.js
jquense: just left over while refactoring oops

src/DropdownToggle.js
jquense: maybe, i'm not convinced the pure hook api here is the best option, at least not until there is a proper popper.js hooks api

src/DropdownMenu.js
jquense: probably...this was the way it was i think

www/gatsby-config.js
jquense: it doesn't work :/

src/RootCloseWrapper.js
```diff
@@ -1,20 +1,89 @@
...
+  { disabled, clickTrigger = 'click' } = {},
+) {
+  const preventMouseRootCloseRef = useRef(false);
+  const onClose = useEventCallback(onRootClose || noop);
+
+  const handleMouseCapture = e => {
+    const currentTarget = ref && ('current' in ref ? ref.current : ref);
+    warning(
+      !!currentTarget,
+      'RootClose captured a close event but does not have a ref to compare it to. ' +
```
taion: i'd just pre-concat these; now that we're on prettier, we won't get a line length lint hit anyway

src/RootCloseWrapper.js
```diff
@@ -1,20 +1,89 @@
...
+    // avoid false positives in the contains() check below if the target DOM
+    // element is removed in the React mouse callback.
+    const removeMouseCaptureListener = listen(
+      document,
+      clickTrigger,
+      handleMouseCapture,
+      true,
+    );
+
+    const removeMouseListener = listen(document, clickTrigger, handleMouse);
```
taion: do we not need `handleMouse` and `handleKeyUp` for exhaustive dep tracking?

src/utils/mergeRefs.js
```diff
@@ -0,0 +1,22 @@
+import { useMemo } from 'react';
```
taion: move to restart/hooks?

jquense:
gonna clean up the lint in a min

package.json
```diff
@@ -51,38 +51,44 @@
...
-    "react-context-toolbox": "^2.0.2",
-    "react-popper": "^1.3.2",
-    "uncontrollable": "^6.0.0",
-    "warning": "^4.0.2"
+    "react-popper": "^1.3.3",
+    "uncontrollable": "^6.1.0",
+    "warning": "^4.0.3"
   },
   "peerDependencies": {
     "react": ">=16.3.0",
```
nihgwu: we should bump up the peer dependencies as we are using hooks now

PR #293, opened 2019-04-02
Fix #292: modal shouldn't make itself aria-hidden

### LINKED ISSUE #292: Bug: modal makes itself aria-hidden
When using a `react-bootstrap` `Modal`, I'm unable to navigate the modal using a screen reader

Inspecting the DOM, I see that the outer `<div role="dialog" aria-modal="true">` element has `aria-hidden="true"` set on it (while the modal is open)

I tracked down the logic to what looks like a simple bug fix in `src/utils/manageAriaHidden.js`. Will submit a fixing PR momentarily

helixbass:
Fixes #292 

Not sure where to add a test for this within this repo but making this change manually in my local project dependency did get rid of the `aria-hidden="true"` on the root `<div role="dialog" aria-modal="true">` element

### DIFF
--- a/src/utils/manageAriaHidden.js
+++ b/src/utils/manageAriaHidden.js
@@ -1,30 +1,30 @@
-const BLACKLIST = ['template', 'script', 'style']
+const BLACKLIST = ['template', 'script', 'style'];
 
 let isHidable = ({ nodeType, tagName }) =>
-  nodeType === 1 && BLACKLIST.indexOf(tagName.toLowerCase()) === -1
+  nodeType === 1 && BLACKLIST.indexOf(tagName.toLowerCase()) === -1;
 
 let siblings = (container, exclude, cb) => {
-  exclude = [].concat(exclude)
-  ;[].forEach.call(container.children, node => {
+  exclude = [].concat(exclude);
+  [].forEach.call(container.children, node => {
     if (exclude.indexOf(node) === -1 && isHidable(node)) {
-      cb(node)
+      cb(node);
     }
-  })
-}
+  });
+};
 
 export function ariaHidden(show, node) {
-  if (!node) return
+  if (!node) return;
   if (show) {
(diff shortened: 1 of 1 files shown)


src/utils/manageAriaHidden.js
```diff
@@ -1,30 +1,30 @@
-const BLACKLIST = ['template', 'script', 'style']
+const BLACKLIST = ['template', 'script', 'style'];
```
helixbass: Most of these changes are just Prettier - apparently now you don't use `semi: false`?

Might behoove you to do a PR where you reformat the whole codebase to match current rules so that PR diffs are more discernible?

src/utils/manageAriaHidden.js
```diff
@@ -1,30 +1,30 @@
...
+    node.setAttribute('aria-hidden', 'true');
   } else {
-    node.removeAttribute('aria-hidden')
+    node.removeAttribute('aria-hidden');
   }
 }
 
-export function hideSiblings(container, { root, backdrop }) {
-  siblings(container, [root, backdrop], node => ariaHidden(true, node))
+export function hideSiblings(container, { dialog: root, backdrop }) {
```
helixbass: This (and the same below for `showSiblings()`) is the only actual change:

When I logged `root` it was `undefined`

Scanning the source code for this repo and `react-bootstrap`, I didn't see any relevant assignments/references to `root`

So it looks like this is a bug, the `dialog` property on `Modal`s (set by `Modal::setDialogRef()`) appears to be the correct element here

chazzlabs:
Is there any intent on merging this fix?  It's a bit concerning that activity on issues and PRs in the repo has ceased since mid-March.

jquense:
sometimes people with full time jobs get busy for periods of time and don't have a lot of extra time to do free work on the internet. Don't be concerned, life just gets busy.

PR #296, opened 2019-06-09
Support esm with package.json routers

TrySound:
Similar to react-transition-group approach. Also added babel runtime with esm helpers.

Closes https://github.com/react-bootstrap/react-overlays/pull/282

### DIFF
--- a/.babelrc.js
+++ b/.babelrc.js
@@ -14,6 +14,7 @@ module.exports = api => ({
   ],
   plugins: [
     ['@babel/plugin-proposal-class-properties', { loose: true }],
+    ['@babel/plugin-transform-runtime', { useESModules: api.env() === 'esm' }],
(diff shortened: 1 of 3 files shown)


TrySound:
Looks like lint errors are unrelated.

taion:
On the runtime thing as well, see https://github.com/reactjs/react-transition-group/pull/507. I think it's a bug in the Babel preset.

TrySound:
Runtime is used for both cjs and esm here.

jquense:
lint errors are expected, it's been in a bad state since #288 i've not quite gotten to it yet

taion:
Ah, I see – you're adding the runtime transform here.

TrySound:
@jquense Lint is fixed. Could you merge?

Issue #283, opened 2019-02-01
Body element should not be required at import time


nip3o:
I'm tying to migrate from react-overlays 0.8.3 to 1.1.1, but after upgrading to the newer version we run into the following crash when our webpack bundle loads:

    TypeError: document.body is null[Learn More] vendors.2966214b0af6eeb1dc74.js line 3642 > eval:21:7
        scrollbarSize webpack:///./node_modules/dom-helpers/util/scrollbarSize.js?:21
    ModalManager2React	js http://localhost:8000/static/assets/system/vendors.2966214b0af6eeb1dc74.js:4644
        __webpack_require__ http://localhost:8000/static/assets/system/bundle.084924a405a4fa2f81d5.js:84
        <anonymous> webpack:///./client/system/components/base/modal/modal.tsx?:14
        tsx http://localhost:8000/static/assets/system/bundle.084924a405a4fa2f81d5.js:1502
        __webpack_require__ http://localhost:8000/static/assets/system/bundle.084924a405a4fa2f81d5.js:84
        <anonymous> webpack:///./client/system.js?:365
        <anonymous> webpack:///./client/system.js?:2
        js http://localhost:8000/static/assets/system/bundle.084924a405a4fa2f81d5.js:922
        __webpack_require__ http://localhost:8000/static/assets/system/bundle.084924a405a4fa2f81d5.js:84
        checkDeferredModules http://localhost:8000/static/assets/system/bundle.084924a405a4fa2f81d5.js:46
        <anonymous> http://localhost:8000/static/assets/system/bundle.084924a405a4fa2f81d5.js:219
        <anonymous> http://localhost:8000/static/assets/system/bundle.084924a405a4fa2f81d5.js:1

This is because the `script` tag that loads our bundle is located in `head`, so `body` does not exist at import time. Since styles and other libs are depending on being loaded before `body`, moving the bundle is not an option. Instead I believe this code should run after the page has finished loading?

taion:
this might be a bug, but you should not in general load your JS bundle until after your page has finished loading anyway

nip3o:
The main reason why we are doing that is since we are using `css-loader` to load our CSS with webpack, so importing the bundle after page load will render a page completely without styling for a short while unless loading the bundle before `body`. I agree that, for example, trying to render a modal before document has loaded would be a misuse, but not just importing the module.

jquense:
[...]

PR #297, opened 2019-06-12
fix: Compute scrollbar size dynamically on need

billychan:
About issue #283 and react bootstrap issue https://github.com/react-bootstrap/react-bootstrap/issues/3645

Verified locally react-bootstrap modal launched well.


billychan:
@taion you are right at that point. How do you like if the manager instance instantiated here? https://github.com/billychan/react-overlays/blob/c885aa3102ec01569d18df447f67cb2f3a950eb9/src/Modal.js#L235

taion:
Yes – I think `componentDidMount` is the right place to construct the default modal manager.

billychan:
@taion Thanks checking man, please see how the new version looks 😃

src/Modal.js
```diff
@@ -233,6 +230,7 @@ class Modal extends React.Component {
 
   componentDidMount() {
     this._isMounted = true;
+    this.manager = new ModalManager();
```
taion: - You should only instantiate this if a manager wasn't specified via `props`.
- By default, we still want a single global singleton manager. We just want to delay when it gets initialized.

So I think I'd do something like:

```js
let manager; // Maybe export this?

class Modal {
  // ...

  getModalManager() {
    if (this.props.manager) {
      return this.props.manager;
    }

    if (!manager) {
      manager = new ModalManager();
    }

    return manager;
  }

  // ...
}
```

billychan: My bad, I previously wrote that as `this.manager = this.props.manager || new ModalManager` to still allow injection from props, removed when testing some issue and forgot to revert. Thanks catching 👍

taion: But even so that doesn't quite preserve the semantics, as it means that changing the passed-in modal manager via the props would have no effect.

billychan: I saw your point, you want to keep a singleton of ModalManager instance which keeps global knowledge of `this.modals`.

That might be better wrapped in a separate module I guess, maybe something like this?

```
// Modal.js
import { getModalManager } from './ModalManager';

//.....
  componentDidMount() {
    this._isMounted = true;
    this.manager = this.props.manager || getModalManager();
```

And

```
// ModalManager.js
// ....

let manager;

function getModalManager() {
  if (manager) {
     return manager;
  }
 manager = new ModalManager();
 return manager;
}
```

Sounds good?

BTW maybe similar needs to be done in react-bootstrap later too https://github.com/react-bootstrap/react-bootstrap/blob/0bb8a6fd4ad110c49e5e89f05228ade0968652bf/src/Modal.js#L154

taion: See the example code in https://github.com/react-bootstrap/react-overlays/pull/297#discussion_r294290387. Do that, then use `this.getModalManager()`. The issue with your proposed solution above is that it wouldn't respect changes to the prop.

deXterbed:
@taion any update on when this PR will be merged? thanks!

src/Modal.js
```diff
@@ -314,6 +313,18 @@ class Modal extends React.Component {
     this.backdrop = ref && ReactDOM.findDOMNode(ref);
   };
 
+  getModalManager = () => {
```
taion:
```suggestion
  getModalManager() {
```

src/Modal.js
```diff
@@ -314,6 +313,18 @@ class Modal extends React.Component {
...
+    if (this.props.manager) {
+      return this.props.manager;
+    }
+
+    if (!manager) {
+      manager = new ModalManager();
+    }
+
+    return manager;
+  };
```
taion:
```suggestion
  }
```

deXterbed:
@billychan can you look into the changes suggested by taion?

src/Modal.js
billychan: right 👍

billychan:
@deXterbed @taion I've merged back master and updated code to remove binding. 

The spec still failed at uploading tests coverage step, guess that's not related.

taion:
Thanks!

PR #299, opened 2019-06-14
Use popper

jquense:
I think done


src/usePopper.js
```diff
@@ -0,0 +1,121 @@
...
+  left: '0',
+  opacity: '0',
+  pointerEvents: 'none',
+};
+
+const initialArrowStyles = {};
+
+const getRefValue = refOrValue =>
+  refOrValue !== null && 'current' in refOrValue
+    ? refOrValue.current
```
taion: can we just require this to be `ref.current`, and not try to be cute here? i'm worried about things possibly going wrong. it doesn't seem like this gets us much DX-wise

src/usePopper.js
```diff
@@ -0,0 +1,121 @@
...
+    positionFixed = false,
+    eventsEnabled = true,
+    modifiers = {},
+  } = {},
+) {
+  const popperInstanceRef = useRef();
+
+  const hasArrow = !!(modifiers.arrow && modifiers.arrow.element);
+
+  const scheduleUpdate = useCallback(() => {
```
taion: this can't ever invalidate, right? maybe just inline this in all cases? or at least there's no point in putting this in state

src/DropdownMenu.js
```diff
@@ -1,12 +1,101 @@
...
+import { useContext, useRef } from 'react';
 
+import useCallbackRef from '@restart/hooks/useCallbackRef';
+import usePopper from './usePopper';
 import useRootClose from './useRootClose';
-import { useMergedRefs } from './utils/mergeRefs';
 import DropdownContext from './DropdownContext';
 
+export function useDropdownMenu(options = {}) {
+  // const prevProps = usePrevious(props);
```
taion: ?

i guess this is all in `usePopper` now?

src/usePopper.js
jquense: we need to allow the straight ref case, b/c it's essentially required for doing updates in response to a ref being set. I do hate this logic tho, maybe we can go the other way and require only callback refs?

taion: What do you mean? The caller of `usePopper` can always pass in an unwrapped ref for the arrow, no?

src/usePopper.js
jquense: The only reason it's in state is b/c we pass it back to the user, and it keeps it stable without an additional memo

src/DropdownMenu.js
jquense: yeah I think it's unneeded here now, but i need to try in RB to confirm

src/usePopper.js
jquense: ooo right this is only for the arrow. hmm yeah maybe, tho it's sort of weird to have an api that only takes a `useRef` ref and not any ref? idk actually, it's odd that there is this distinction

taion: i'm suggesting we kill `getRefValue` and force the caller to pass in e.g. `arrowElementRef.current`

src/Dropdown.js
```diff
@@ -1,11 +1,12 @@
 import matches from 'dom-helpers/query/matches';
 import qsa from 'dom-helpers/query/querySelectorAll';
-import React from 'react';
-import ReactDOM from 'react-dom';
+import React, { useCallback, useRef, useEffect, useMemo } from 'react';
 import PropTypes from 'prop-types';
-import uncontrollable from 'uncontrollable';
+import useUncontrolled from 'uncontrollable/hook';
```
TrySound: Upgrade uncontrollable

TrySound:
Hi @jquense do you need some help with this?

jquense:
happy to get some help if you wanna jump in I'm just a bit short on time at the moment

bpas247:
Should we close this for now? Since it’s being superseded by #308

src/Dropdown.js
```diff
@@ -88,82 +95,130 @@ const defaultProps = {
...
+      toggleElement.focus();
+    }
+  });
 
-  componentDidUpdate(prevProps) {
-    const { show } = this.props;
-    const prevOpen = prevProps.show;
+  const maybeFocusFirst = useEventCallback(() => {
+    const type = lastSourceEvent.current;
+    let focusStype = focusFirstItemOnShow;
```
taion:
```suggestion
    let focusType = focusFirstItemOnShow;
```

src/Dropdown.js
```diff
@@ -88,82 +95,130 @@ const defaultProps = {
...
 
-    if (show && !prevOpen) {
-      this.maybeFocusFirst();
+    if (focusStype == null) {
```
taion: Does it make sense to `useMemo` this outside of this callback? I don't think this needs to be calculated on-the-fly.

src/Dropdown.js
```diff
@@ -88,82 +95,130 @@ const defaultProps = {
...
+    if (first && first.focus) first.focus();
+  });
+
+  useEffect(() => {
+    if (show) maybeFocusFirst();
+    else if (focusInDropdown.current) {
+      focusInDropdown.current = false;
+      focusToggle();
+    }
+    // only `show` should be changing
```
taion: Does it make sense to guard and check the previous value of `show` anyway?

src/DropdownMenu.js
```diff
@@ -1,12 +1,99 @@
...
+    context.alignEnd == null ? options.alignEnd : context.alignEnd;
+
+  if (show && !hasShownRef.current) {
+    hasShownRef.current = true;
+  }
+
+  const handleClose = e => {
+    if (!context.toggle) return;
+
+    context.toggle(false, e);
```
taion:
```suggestion
    context.toggle?.(false, e);
```

src/DropdownMenu.js
```diff
@@ -1,12 +1,99 @@
...
+  const show = context.show == null ? options.show : context.show;
+  const alignEnd =
+    context.alignEnd == null ? options.alignEnd : context.alignEnd;
+
+  if (show && !hasShownRef.current) {
+    hasShownRef.current = true;
+  }
+
+  const handleClose = e => {
+    if (!context.toggle) return;
```
taion:
```suggestion
```

src/DropdownMenu.js
```diff
@@ -1,12 +1,99 @@
...
+
```
taion:
```suggestion
```

src/DropdownMenu.js
```diff
@@ -1,12 +1,99 @@
...
+    context.toggle(false, e);
+  };
+
+  const { drop, setMenu, menuElement, toggleElement } = context;
+
+  let placement = alignEnd ? 'bottom-end' : 'bottom-start';
+  if (drop === 'up') placement = alignEnd ? 'top-end' : 'top-start';
+  if (drop === 'right') placement = alignEnd ? 'right-end' : 'right-start';
```
taion:
```suggestion
  elif (drop === 'right') placement = alignEnd ? 'right-end' : 'right-start';
```

src/DropdownMenu.js
```diff
@@ -1,12 +1,99 @@
...
+  if (drop === 'left') placement = alignEnd ? 'left-end' : 'left-start';
```
taion:
```suggestion
  elif (drop === 'left') placement = alignEnd ? 'left-end' : 'left-start';
```

src/DropdownMenu.js
TrySound: Is this a javascript?

taion: lol

taion: it's amazing what you can do with babel these days :p

src/DropdownMenu.js
TrySound: Optional chaining plugin is not specified in babel config

src/Dropdown.js
jquense: in practice it only runs on show so it's probably fine

TrySound:
Is this ready to merge guys?

PR #302, opened 2019-07-03
Install www dependencies in install hook

TrySound:
The problem with broken build are not installed dependencies in www.
As a solution I suggest to run yarn for www after installing
dependencies on CI.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,7 +16,7 @@ before_install:
   - export DISPLAY=:99.0
   - sh -e /etc/init.d/xvfb start
 
-after_install:
+before_script:
   - cd www
   - yarn
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,9 +16,8 @@ before_install:
   - export DISPLAY=:99.0
   - sh -e /etc/init.d/xvfb start
 
-before_script:
-  - cd www
-  - yarn
+install:
+  - yarn --cwd www
 
 after_script:
   - node_modules/.bin/codecov
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,7 @@ sudo: false
 
 language: node_js
 node_js:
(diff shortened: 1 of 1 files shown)


mxschmitt:
Which broken builds? @TrySound

TrySound:
image
image

.travis.yml
```diff
@@ -2,7 +2,7 @@ sudo: false
 
 language: node_js
 node_js:
-  - stable
+  - node
```
mxschmitt: why do you change this?

.travis.yml
```diff
@@ -16,6 +16,10 @@ before_install:
   - export DISPLAY=:99.0
   - sh -e /etc/init.d/xvfb start
 
+before_script:
```
mxschmitt:
```suggestion
install:
```

.travis.yml
```diff
@@ -16,6 +16,10 @@ before_install:
...
+  - cd www
+  - yarn
```
mxschmitt:
```suggestion
  - yarn --cwd www
```

.travis.yml
TrySound: stable was deprecated long time ago in favor of node

TrySound: https://docs.travis-ci.com/user/languages/javascript-with-nodejs/#specifying-nodejs-versions

TrySound:
Yay, we did it

.travis.yml
mxschmitt: nice!

mxschmitt:
@TrySound seems not to be fixed :/ 
https://travis-ci.org/react-bootstrap/react-overlays/builds/553709192#L464

TrySound:
Weird, looks like install hook overwrote normal `yarn` call. But somehow PR worked.

PR #303, opened 2019-07-03
fix: travis

mxschmitt:
I hade no motivation to use nice commit messages. And we squash either.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,9 +7,11 @@ node_js:
 env:
   - BROWSER=ChromeCi
 
+dist: xenial
+
 cache:
-  directories:
-    - node_modules
+  - npm
+  - yarn
 
 before_install:
   - export CHROME_BIN=chromium-browser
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -8,15 +8,16 @@ env:
   - BROWSER=ChromeCi
 
 cache:
-  directories:
-    - node_modules
+  - npm
```
TrySound: Why npm?

mxschmitt: node_modules

taion: using `yarn` alone should be sufficient then?

mxschmitt: unfortunately both are required: https://docs.travis-ci.com/user/caching/#yarn-cache

taion: yarn actually just takes everything from its cache dir anyway (instead of re-downloading). it's the same w/how travis doesn't cache `node_modules` when using `npm ci`

PR #306, opened 2019-07-03
Cleanup package dependencies

TrySound:
- upgraded hooks and uncontrollable with esm support
- removed not used classnames and prop-types-extra

### DIFF
--- a/package.json
+++ b/package.json
@@ -51,13 +51,11 @@
     "conventionalCommits": true
   },
   "dependencies": {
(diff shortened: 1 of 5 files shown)


TrySound:
Are you sure global in eslint config is related to classnames package?

mxschmitt:
> Are you sure global in eslint config is related to classnames package?

nope, not totally sure.

TrySound:
Looks like it's not used anyway.

PR #311, opened 2019-07-15
chore: Update dependencies

taion:


taion:
Not touching react-live for now because it's actually a substantial change. Otherwise doing this ahead of turning on Dependabot so we don't get spammed.

PR #372, opened 2019-08-25
ci: add github actions and integrate semantic-release

bpas247:
Relevant to 4324 in the React-Bootstrap repository.

Adds github actions, with a workflow for running tests and a
workflow for release.

For the release workflow, there is an integration with
semantic-release. We would need to need to setup two github
secrets for semantic to work properly: `NPM_TOKEN` and `GH_TOKEN`.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,20 @@
+name: Release New Version
+on:
+  push:
+    branches:
+      - master
+jobs:
+  release:
+    steps:
+      - name: Checking out Repository
+        uses: actions/checkout@v1
+      - name: Setup Node.js
+        uses: actions/setup-node@v1
+        with:
+          node-version: 12.x
+      - name: Install Dependencies
+        run: yarn bootstrap
+      - name: Build Production Files
+        run: yarn build
+      - name: Release To NPM and GitHub
+        run: npx semantic-release
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,0 +1,29 @@
+name: Run Tests
+on: [push]
+jobs:
+  test:
+    runs-on: ${{ matrix.os }}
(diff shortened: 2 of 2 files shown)


mxschmitt:
I would not use the codecov Action, because we then fetch the dependencies multiple times and rum the tests multiple times. The codecov npm package works great :)

bpas247:
> I would not use the codecov Action, because we then fetch the dependencies multiple times and rum the tests multiple times. The codecov npm package works great :)

Sure, but if the concern is that it will upload the test coverage multiple times, then it could just be solved by having it run as a separate job (which is how I organized it here, as it re-runs the tests and uploads the coverage results after the main test job completes with a success).

mxschmitt:
I would still not use the codecov action. We only need to execute `npx codecov` and it will be uploaded. No need for spawning another Docker container (https://github.com/codecov/codecov-action/blob/master/Dockerfile). Both are doing actually the same except, that in the Docker container the bash script of codecov will be ran, which detects the language etc.

mxschmitt:
> > I would still not use the codecov action. We only need to execute npx codecov and it will be uploaded. No need for spawning another Docker container (https://github.com/codecov/codecov-action/blob/master/Dockerfile). Both are doing actually the same except, that in the Docker container the bash script of codecov will be ran, which detects the language etc. In the end the codecov Action has more pitfalls.
> 
> So I'm gonna have to disagree with you, because if you look at the CI [build logs](https://github.com/react-bootstrap/react-bootstrap/pull/4324/checks?check_run_id=202655374) for the react-bootstrap PR then you'd see that even though the code coverage is failed to be reported, the CI build still passes. Using the GitHub action in this PR actually catches the CI build failing when it's not able to upload the report.

True, I've submitted a fix for it here: https://github.com/codecov/codecov-node/pull/144

Issue #240, opened 2018-03-27
ariaHidden(false, ..) should (probably) use modalNode insted of mountNode


ruslansavenok:
`ariaHidden` is added to modal div, not it's container.

But this line tries to remove it from container.
https://github.com/react-bootstrap/react-overlays/blob/4498bb1ce4375e72433e9d90a686b37e06a3844b/src/ModalManager.js#L141

omniverse:
Yes! This is messing up screen readers.

omniverse:
Actually, the code above is obsolete. But there are still issues with aria-hidde="true" getting set on the dialog itself.

The hideSiblings param declaration `hideSiblings(container, { root, backdrop })` doesn't match what is being passed in.
On https://github.com/react-bootstrap/react-overlays/blob/master/src/ModalManager.js#L93 should be something like `      hideSiblings(container, {root: modal.dialog}); `

cameron-martin:
> Actually, the code above is obsolete.

This is the code in the latest stable version. Would you consider releasing a patch for this?

PR #389, opened 2019-08-27
Mismatch between Modal.dialog and hideSiblings parameters

omniverse:
Fixes the issue raised here: https://github.com/react-bootstrap/react-overlays/issues/240

The modal component contains refs to `dialog` and `backdrop` https://github.com/omniverse/react-overlays/blob/master/src/Modal.js#L309-L315
not `root` and `backdrop` as destructured in manageAriaHidden https://github.com/omniverse/react-overlays/blob/master/src/Modal.js#L309-L315

So the aria-hidden="true" attribute was getting applied to the dialog itself and breaking screen readers, accessibility features, etc.

### DIFF
--- a/src/utils/manageAriaHidden.js
+++ b/src/utils/manageAriaHidden.js
@@ -21,10 +21,10 @@ export function ariaHidden(show, node) {
   }
 }
 
-export function hideSiblings(container, { root, backdrop }) {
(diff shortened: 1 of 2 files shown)


omniverse:
> Nice catch!
> 
> Is there any chance you could add a regression test here?

Added

jquense:
i don't understand how to read these CI checks...where is even the logs for CI?

PR #398, opened 2019-08-28
ci(actions): add deploy documentation action

bpas247:
This should allow us to not having to worry about the chore task of building and deploying the docs manually.

### DIFF
--- a/.github/workflows/check-docs.yml
+++ b/.github/workflows/check-docs.yml
@@ -10,8 +10,7 @@ jobs:
       - name: Setup Node.js
         uses: actions/setup-node@v1
       - name: Install Dependencies
-        working-directory: www
-        run: yarn install
+        run: yarn bootstrap
       - name: Build Production Files
         working-directory: www
         run: yarn build
--- a/.github/workflows/check-docs.yml
+++ b/.github/workflows/check-docs.yml
@@ -1,0 +1,17 @@
+name: Check Docs
+on: [push]
+jobs:
+  build_docs:
+    name: Build Doc Production Files
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checking out Repository
+        uses: actions/checkout@v1
+      - name: Setup Node.js
+        uses: actions/setup-node@v1
+      - name: Install Dependencies
+        working-directory: www
+        run: yarn install
+      - name: Build Production Files
+        working-directory: www
+        run: yarn build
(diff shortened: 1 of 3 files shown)


mxschmitt:
> Eh, I feel like this action isn’t that useful due to the netlify integration. It might be more useful if it also ran the deployment script, and only ran this action on pushes to master.

yep! I think Netlify is enough. We only need an action which deploys on master to GitHub pages. I already started for react-bootstrap.

```
name: Build and Deploy documentation
on:
  push:
    branches:
    - master
jobs:
  build:
    runs-on: ubuntu-latest
    steps:
    - uses: actions/checkout@v1
    - name: Setup Node.js
      uses: actions/setup-node@v1
      with:
        node-version: 12.x
    - name: Installing Dependencies
      run: yarn bootstrap
    - name: Build documentatation
      run: yarn --cwd www build
    - name: Publish documentation to GitHub Pages
      run: |
        cd www
        git init
        git remote add origin https://github.com/react-bootstrap/react-bootstrap.github.io.git
        git add .
        git commit -m "Updated documentation by GitHub Actions"
        git push origin master --force
```

Something like that we need here but instead deploying to the master branch, we need to deploy to the gh-pages branch.

bpas247:
> Something like that we need here but instead deploying to the master branch, we need to deploy to the gh-pages branch.

or just have it run the [deploy-docs](https://github.com/react-bootstrap/react-overlays/blob/39dc831822d45428cfa69c3791a48a8022c340b7/package.json#L30) npm script :)

bpas247:
This probably won't work until the build is fixed, so I'll wait to merge this until after the fix has been pushed. I tried tracking down the issue on several occasions, but to no avail.

I think it might be related to importing `dom-helpers` module that utilizes `document`, which isn't an issue in a development enviroment but does cause [issues when building for production](https://www.gatsbyjs.org/docs/debugging-html-builds/).

bpas247:
> Looks good, but I guess we need to provide authentication for GitHub in the runner itself.

Like an access token? GitHub [provides one by default](https://help.github.com/en/articles/virtual-environments-for-github-actions#github_token-secret), we just need to bring it into the environment of the deploy step.

bpas247:
hmm, it didn't seem to like that configuration so I might go with @mxschmitt's [original suggestion](https://github.com/react-bootstrap/react-overlays/pull/398#issuecomment-525627193) in a new PR.

bpas247:
We could also go with an [action that does this for us](https://github.com/JamesIves/github-pages-deploy-action), so I'm currently researching options for what to do next.

PR #412, opened 2019-08-30
fix(docs): compilation issues when building production files

bpas247:
This was due to `ownerDocument` using `document` when it wasn't
defined (which is the case when building production files).


src/utils/useWaitForDOMRef.js
```diff
@@ -2,6 +2,7 @@ import ownerDocument from 'dom-helpers/ownerDocument';
 import { useState, useEffect } from 'react';
 
 const resolveRef = ref => {
+  if(typeof document === 'undefined') return undefined;
```
bpas247: So I'm pretty sure returning `undefined` here isn't the correct behavior for this use case, does anyone have a better suggestion for what to return here?

PR #430, opened 2019-09-02
fix(actions): migrate deploy docs step to use external action

bpas247:
Using an external action here saves us from setting this up
manually.

also removes an uneeded dev dependency, as we aren't deploying
the docs locally anymore.


.github/workflows/deploy-docs.yml
```diff
@@ -15,6 +15,9 @@ jobs:
       - name: Install Dependencies
         run: yarn bootstrap
       - name: Build and Deploy Production Files
+        uses: JamesIves/github-pages-deploy-action@master
         env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-        run: yarn deploy-docs
+          ACCESS_TOKEN: ${{ secrets.ACCESS_TOKEN }}
```
bpas247: So this is unfortunate, but until [server-to-server requests kick off page builds](https://github.com/JamesIves/github-pages-deploy-action/issues/5#issuecomment-476224937), there isn't a good workaround for this.

We have a couple options here:
- Wait until this issue is resolved and the subsequent [PR](https://github.com/JamesIves/github-pages-deploy-action/pull/13) is merged.
- Create a [machine user](https://developer.github.com/v3/guides/managing-deploy-keys/#machine-users) and create a new secret out of that. 

IMO the former seems like the less painful route to go.

taion:
Wait, if https://github.com/JamesIves/github-pages-deploy-action/pull/13 was merged, do we still need that token?

PR #439, opened 2019-09-04
ci(actions): improve alignment of test configuration with React-Bootstrap

bpas247:
closes #179

PR #446, opened 2019-09-04
Upgrade dom-helpers

TrySound:
Ref https://github.com/react-bootstrap/dom-helpers/commit/87e2fe448b5e852574e1ee63bd3f2102b4c65d79

/cc @jquense

### DIFF
--- a/package.json
+++ b/package.json
@@ -58,7 +58,7 @@
   "dependencies": {
     "@babel/runtime": "^7.4.5",
     "@restart/hooks": "^0.3.12",
-    "dom-helpers": "^3.4.0",
+    "dom-helpers": "^5.0.1",
     "popper.js": "^1.15.0",
     "prop-types": "^15.7.2",
     "uncontrollable": "^7.0.0",
--- a/package.json
+++ b/package.json
@@ -58,7 +58,7 @@
   "dependencies": {
     "@babel/runtime": "^7.4.5",
     "@restart/hooks": "^0.3.12",
-    "dom-helpers": "^5.0.1",
+    "dom-helpers": "^5.1.0",
     "popper.js": "^1.15.0",
     "prop-types": "^15.7.2",
     "uncontrollable": "^7.0.0",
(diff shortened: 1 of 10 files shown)


taion:
might want to update RTG first, though, since we're still pulling in dom-helpers via RTG

TrySound:
Done @taion https://github.com/reactjs/react-transition-group/pull/549

bpas247:
hmm, strange that the CI build didn't kick off. Is this branch up to date with the master branch?

TrySound:
Yes, I pulled master before submmitting.

bpas247:
I just ran the build locally, and it seems to be failing on [this test](https://github.com/react-bootstrap/react-overlays/blob/e8debc2e6814397c6bfc26299b2117b3603ca1ef/test/ModalSpec.js#L416-L434) with the following error:
```
Error: Uncaught AssertionError: expected 'focus-container' to include 'modal' (node_modules/chai/chai.js:239)
```

TrySound:
@bpas247 Looks like it was race condition.

jquense:
yeah that happens occasionally when browser windows aren't actively focused

bpas247:
We've changed up the CI configuration since this PR has been opened (see #448 and #439), so we'll probably need to merge the changes from master into this branch to get those required status checks to kick off properly.

@TrySound do you mind doing that? If not, I'd be more than happy to 😃

taion:
I merged this to master. Will merge once CI goes green.

taion:
okay... so the way we have the codecov secret set up means it's not accessible for builds from forks. bleh.

bpas247:
> okay... so the way we have the codecov secret set up means it's not accessible for builds from forks. bleh.

Any easy fix for that?

taion:
I don't see a way in the secret config to expose secrets to builds from forks. We might just have to wait until there's better first-party support here, and just deal with the pain for now. Not a problem in this case anyway because we don't care about coverage diffs for this change.

bpas247:
This seems related to the overall issue of not allowing [token-free uploads for public repos](https://github.com/codecov/codecov-action/issues/11), so hopefully they'll push a fix for it soon.

PR #448, opened 2019-09-04
ci(actions): fix for build not triggering on forked pull request

bpas247:
Currently, the build is only being triggered whenever there is
a push to the repository.

This does not lead to a build being triggered whenever there is a
push to a branch on a forked repo.

This fix allows it to trigger on every push *or* every pull
request.

We'll probably need to also fix the configuration on the
React-Bootstrap repo.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,5 +1,11 @@
 name: Run Tests
-on: [push, pull_request]
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+    branches:
+      - master
 jobs:
   test:
     name: Run Testing Suite
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,5 +1,5 @@
 name: Run Tests
-on: [push]
+on: [push, pull_request]
 jobs:
   test:
     name: Run Testing Suite


bpas247:
Eh, Idk if I like this setup either. Maybe we can separate the pull request job into its own workflow, with a more simple setup for quick testing? Then we can have the push job to be more comprehensive.

taion:
Does this work roughly like the setup for Travis?

Ideally we'd only want to automatically run tests on PRs to master, maybe.

bpas247:
> Does this work roughly like the setup for Travis?

Yep, it's essentially the same AFAIK 👍

bpas247:
> i'd consider removing push. either way

hmm, don't we still want our CI builds to trigger when there is a push to master though?

taion:
ah ok, my bad

i was thinking the right setup would be pushes and PRs to master, but not pushes to other branches (unless there's a PR to master open)

bpas247:
> i was thinking the right setup would be pushes and PRs to master, but not pushes to other branches (unless there's a PR to master open)

I think I might be able to set this up in a map event configuration actually, I'll tinker with it and see what I can come up with 👌

bpas247:
@taion do you mind checking this configuration? I just wanted to confirm that this is what you were thinking

PR #488, opened 2019-09-15
Drop warning

TrySound:
I tend to remove warning because of confusing api. It's hard to say by
which condition warning is shown. This package also requires babel
plugin setup which is forgotted here. IMO it's much easier to just write
console statement and choose either console.warn or console.error than
dealing with this commonjs-only package.

### DIFF
--- a/package.json
+++ b/package.json
@@ -61,8 +61,7 @@
     "dom-helpers": "^5.1.0",
     "popper.js": "^1.15.0",
     "prop-types": "^15.7.2",
-    "uncontrollable": "^7.0.0",
-    "warning": "^4.0.3"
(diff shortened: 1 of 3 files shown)


TrySound:
Hi @jquense what do you think about this?

taion:
eh... i'm not that bothered by the API. it's also more or less ubiquitous at this point.

@TrySound  would you be happier if we switched to `tiny-warning`?

TrySound:
I'm not really a fan of this convention anymore caz it's confused me a few times lately. Also tiny-warning uses console.warn which is kinda a breaking change. Though it's not me to decide :)

PR #559, opened 2019-10-02
docs: re-do a bunch of the docs

jquense:


jquense:
prepping for a proper v2 release

www/src/components/Anchor.js
```diff
@@ -0,0 +1,20 @@
+import PropTypes from 'prop-types';
+import React from 'react';
+
+export default class Anchor extends React.Component {
+  static propTypes = {
+    id: PropTypes.string,
+  };
+
+  render() {
+    let id =
```
taion: const?

why isn't this a fn component

www/src/pages/index.js
```diff
@@ -1,246 +1,5 @@
...
-            </p>
-
-            <Playground codeText={TransitionSource} />
-          </section>
-        </main>
-      </div>
-    );
-  }
+function App() {
+  return null;
```
taion: ?

taion: what's the point of this?

jquense: it still gets rendered when running locally sometimes, be we redirect immediately

www/src/components/Anchor.js
jquense: I just copy/pasted it from index.js

PR #597, opened 2019-10-09
React 17 support

MatthewHerbst:
This is to support https://github.com/react-bootstrap/react-bootstrap/issues/4398

This is a breaking change, so would likely need the `0.9.0` tag.

While at it I also upgraded dependencies to their newest minor/patch versions.


MatthewHerbst:
^bump @taion If we did this then I don't think https://github.com/react-bootstrap/react-bootstrap/pull/4639 would be necessary (for now)

ElScorxcho:
@MatthewHerbst Out of curiosity, what are the breaking changes?

MatthewHerbst:
@ElScorxcho correct, the `peerDependency` change is the breaking change, since `UNSAFE_` methods were not introduced until React 16.3.0

wheeler:
Super excited for this! I was trying out `0.9.0` and it seems like one lifecycle got missed. I prepared PR #682 to fix. cc @taion

taion:
merged; will cut a new R-O release tomorrow

technically don't need a new release here

PR #682, opened 2019-10-31
fix(0.x): missed UNSAFE lifecycle in Portal (React 17)

wheeler:
#597 missed this lifecycle method.


taion:
not sure why we don't just do that in the ctor, but it doesn't really matter

taion:
v0.9.1

wheeler:
@taion Thanks so much for the quick turnaround on these! So many less warnings 🎉

Issue #748, opened 2019-12-09
It is need to know when menu will be rendered and had position


Viktor-the-great:
first of all sorry for my english

Input field is inside menu (Dropdown.Menu). There is a lot of text in front of the input field. Set focus on input field when menu is mounted. Menu has not yet received a position, so the page is scroll to the beginning (menu is hidden and is at the very top of the page).

I am now using timeout to fix it.
In [demo](https://codesandbox.io/s/react-overlays-bug-z1dsi) scroll to bottom. There are two input button to open dropdown. Page scrolls up when first button is pressed. Second button use setTimeout to wait when menu get position.

I suggest using onCreate, onUpdate of Popper.js to solve this problem, and for this to pass these functions to the constructor (usePopper hook)

taion:
Pass in `popperConfig` to set detailed Popper options

Viktor-the-great:
> Pass in popperConfig to set detailed Popper options

I see only popperConfig.modifiers
https://github.com/react-bootstrap/react-overlays/blob/3db0db2b569f86e71c42a0597a1973e8f58144d3/src/DropdownMenu.js#L42-L55

taion:
ah, okay. this should spread in `...popperConfig` as with `<Overlay>` then. can you send in a PR?

Viktor-the-great:
yes I can

taion:
thanks!

PR #750, opened 2019-12-09
pass whole popperConfig to PopperJS (#748)

Viktor-the-great:


src/usePopper.js
```diff
@@ -81,6 +85,7 @@ export default function usePopper(
     };
 
     popperInstanceRef.current = new PopperJS(referenceElement, popperElement, {
+      ...options,
```
taion: should this use an `options.popperConfig`? otherwise we'd need to spread out the properties under `options` that we consume above.

probably doing an `options.popperConfig` makes the typing easier, too

Viktor-the-great: options.placement, options.modifiers, options.eventsEnabled is separate props, options.popperConfig is composite property of the remaining options of Popperjs. I think this could confuse users.

taion: This is how e.g. `<Overlay>` works, though, no?

taion: shoot, my bad, that's _not_ how `<Overlay>` works. it _does_ spread in the popper config like you had here. sorry about that.

but if i'm not mistaken, then, i think in `usePopper`, `...popperConfig` should just be the options _not_ explicitly consumed?

Viktor-the-great: I didn't notice it.

taion: no, sorry, i meant your initial implementation was correct. my suggestion above was bad. we _should_ spread in `...popperOptions` into `usePopper`. we just need to also spread _out_ the options that we explicitly consume there. otherwise it would be a breaking change

taion:
@jquense ping

taion:
@Viktor-the-great sorry for the delay

Issue #746, opened 2019-12-06
restore usage dom-helper.ownerDocument in useRootClose like in RootCloseWrapper


Viktor-the-great:
## Describe the bug

Dropdown inside iframe is not trigger "root close" for iframe document

## To Reproduce

1. Render Dropdown inside iframe
2. Open Menu
3. Click outside Menu (inside iframe)
4. Menu is not closed

## Expected behavior

Dropdown should closed

taion:
Is this a regression?

Viktor-the-great:
If regression is to restore functionality from v1.2.0, then yes

taion:
Regression from v1.2.0, I mean – in which case we'd be happy to take a PR here.

Viktor-the-great:
PR #751

PR #751, opened 2019-12-10
use ownerDocument to get ref to parent document (#746)

Viktor-the-great:


src/useRootClose.js
```diff
@@ -69,23 +71,25 @@ function useRootClose(
   useEffect(() => {
     if (disabled || ref == null) return undefined;
 
+    const doc = ownerDocument(ref && ref.current);
```
taion:
```suggestion
    const doc = ownerDocument(ref.current);
```

we already check that ref is defined above

PR #766, opened 2020-01-22
feat: upgrade popper to v2

jquense:
shooting for a non-breaking upgrade, nice bundle size wins here


jquense:
@FezVrasta awesome work on this version!

www/src/examples/Overlay.js
```diff
@@ -130,18 +130,20 @@ function OverlayExample() {
...
-              placement={placement}
-              style={arrowProps.style}
-            />
-            <Body>
-              I&rsquo;m placed to the <strong>{placement}</strong>
-            </Body>
-          </Tooltip>
-        )}
+        {({ props, arrowProps, placement }) => {
+          return (
```
taion: go back to using shorthand as before?

src/Overlay.js
```diff
@@ -1,10 +1,10 @@
-import PopperJS from 'popper.js';
 import PropTypes from 'prop-types';
 import React, { useState } from 'react';
 import ReactDOM from 'react-dom';
+import { placements } from '@popperjs/core/lib/enums';
```
taion: ouch – no easier import?

src/usePopper.js
```diff
@@ -29,28 +71,58 @@ export default function usePopper(
...
+    }),
+  );
+
+  const updateModifier = useMemo(
+    () => ({
+      name: 'updateStateModifier',
+      enabled: true,
+      phase: 'afterWrite',
+      requires: ['computeStyles'],
+      fn(data) {
```
taion:
```suggestion
      fn: data => {
```

taion:
maybe fix codecov though?

test/DropdownSpec.js
```diff
@@ -352,4 +361,38 @@ describe('<Dropdown>', () => {
       // hence no assert on the input having focus.
     });
   });
+  describe('popper config', () => {
```
taion:
```suggestion

  describe('popper config', () => {
```

src/usePopper.js
```diff
@@ -29,28 +71,58 @@ export default function usePopper(
...
+  );
+  const modifiers = toModifierArray(options.modifiers);
+
+  let eventsModifier = modifiers.find(m => m.name === 'eventListeners');
+  if (!eventsModifier && eventsEnabled) {
+    eventsModifier = {
+      name: 'eventListeners',
+      enabled: true,
+    };
+    modifiers.push(eventsModifier);
```
taion: this is slightly odd in that it may mutate `options.modifiers` (if it was an array), which seems odd.

instead of `toModifierMap`, &c., you might be better off adding some sort of `mergeModifiers` abstraction that's guaranteed to never mutate the input

src/usePopper.js
```diff
@@ -29,28 +71,58 @@ export default function usePopper(
   {
     enabled = true,
     placement = 'bottom',
-    positionFixed = false,
+    strategy = 'absolute',
     eventsEnabled = true,
-    modifiers = {},
+    ...options
```
taion: what's the point of spreading out `...options` if you ignore all the ones other than `modifiers`?

src/usePopper.js
```diff
@@ -29,28 +71,58 @@ export default function usePopper(
...
+          placement: data.state.placement,
+          styles: { ...data.state.styles?.popper },
+          arrowStyles: { ...data.state.styles?.arrow },
+          state: data.state,
+        });
+      },
+    }),
+    [scheduleUpdate, setState],
+  );
+  const modifiers = toModifierArray(options.modifiers);
```
taion:
```suggestion

  const modifiers = toModifierArray(options.modifiers);
```

and should this be `useMemoed`?

taion: including with the "add `eventListeners`" thing

src/usePopper.js
```diff
@@ -60,68 +132,43 @@ export default function usePopper(
...
-    }
-  }, [eventsEnabled]);
+    if (!popperInstanceRef.current || !enabled) return;
+
+    popperInstanceRef.current.setOptions({
+      placement,
+      strategy,
+      modifiers: [...modifiers, updateModifier],
+    });
+    // intentionally NOT re-running on new modifiers
```
taion: why not?

taion:
this basically looks fine; largely implementation notes above

src/usePopper.js
jquense: its very hard to get the memo deps right

jquense: like you'd have to require users memoize `modifiers` i think

PR #775, opened 2020-02-24
Migrate to import * as React from 'react'

TrySound:
Ref https://github.com/facebook/react/pull/18102


jquense:
does this make sense to change until they actually provide an esm build?

TrySound:
Rollup optimise `import * as React from 'react'` into `import{createElement as a}from 'react';` which is not the case with default export prop access which may be side effectful.

Also it's a good for users to see these changes and start migration earlier.

arpit1010:
These changes are needed, prod build is going into error 
`ERROR in [at-loader] ./node_modules/react-overlays/esm/Modal.d.ts:1:8 
    TS1259: Module '"/Users/arpitsharma/mInsights/mplatform-ui-activation-mgmt/node_modules/@types/react/index"' can only be default-imported using the 'esModuleInterop' flag

ERROR in [at-loader] ./node_modules/react-overlays/esm/Overlay.d.ts:1:8 
    TS1259: Module '"/Users/arpitsharma/mInsights/mplatform-ui-activation-mgmt/node_modules/@types/react/index"' can only be default-imported using the 'esModuleInterop' flag
npm ERR! code ELIFECYCLE`

Can you please tell when it will be released ? using 'esModuleInterop' is not an option for me because other modules in my project are not default imported and some issues occur in them

PR #781, opened 2020-03-06
fix: popper imports for common js

### LINKED ISSUE #780: Enable SSR rendering again
With Version 3, I am running into the issue that it seems to no longer work in a SSR scenario. I am probably overlooking something obvious.

I am getting this message:

```
/Users/michael/Development/mine/foobar/node_modules/@popperjs/core/lib/modifiers/arrow.js:1
import getBasePlacement from "../utils/getBasePlacement.js";
^^^^^^

SyntaxError: Cannot use import statement outside a module
    at wrapSafe (internal/modules/cjs/loader.js:1063:16)
    at Module._compile (internal/modules/cjs/loader.js:1111:27)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1167:10)
    at Module.load (internal/modules/cjs/loader.js:996:32)
    at Function.Module._load (internal/modules/cjs/loader.js:896:14)
    at Module.require (internal/modules/cjs/loader.js:1036:19)
    at require (internal/modules/cjs/helpers.js:72:18)
    at Object.<anonymous> (/Users/michael/Development/mine/foobar/packages/foobar/node_modules/react-overlays/cjs/usePopper.js:14:37)
    at Module._compile (internal/modules/cjs/loader.js:1147:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1167:10)
```

That is, this seems to be a problem importing `popper.js` via node. Here is what I think is happening:

`react-overlays/cjs/usePopper.js` tries to require 

```
var _eventListeners = _interopRequireDefault(require("@popperjs/core/lib/modifiers/eventListeners"));
```

However, `@popperjs/core/lib` actually uses `import` syntax, and it's cjs files are in `dist/cjs`.

jquense:
This bundles our slightly smaller popper config for CJS, I think this is fine since the only place this is likely to be used is in a node setting so the possible duplicated popper is probably ok

closes #780


package.json
```diff
@@ -84,6 +84,7 @@
     "@babel/preset-react": "^7.8.3",
     "@react-bootstrap/eslint-config": "^1.3.2",
     "babel-eslint": "^10.1.0",
+    "@rollup/plugin-node-resolve": "^7.1.1",
```
taion:
```suggestion
    "@react-bootstrap/eslint-config": "^1.3.2",
    "@rollup/plugin-node-resolve": "^7.1.1",
    "babel-eslint": "^10.1.0",
```

src/popper.js
```diff
@@ -0,0 +1,23 @@
...
+import computeStyles from '@popperjs/core/lib/modifiers/computeStyles';
+import eventListeners from '@popperjs/core/lib/modifiers/eventListeners';
+import flip from '@popperjs/core/lib/modifiers/flip';
+import hide from '@popperjs/core/lib/modifiers/hide';
+import popperOffsets from '@popperjs/core/lib/modifiers/popperOffsets';
+import preventOverflow from '@popperjs/core/lib/modifiers/preventOverflow';
+import { placements } from '@popperjs/core/lib/enums';
+import { popperGenerator } from '@popperjs/core/lib/popper-base';
+
+export const createPopper = popperGenerator({
```
taion: can you add a comment explaining why we're doing this?

jquense:
```suggestion
// For the common JS build we will turn this file into a bundle with no imports.
// This is b/c the Popper lib is all esm files, and would break in a common js only environment
export const createPopper = popperGenerator({
```

jquense:
Popper has CJS entry but it's for the whole or lite versions. you can't create your own popper from it.

PR #801, opened 2020-04-20
remove `esm` import path

evantahler:
These 2 imports source the `esm` directory of `@restart/hooks` rather than letting the compiler sort out what to do.  This can break things when running in non-typescript environments, or various webpack build steps. 

This bug was found specifically when using react-bootstrap with next.js and running `next build`


evantahler:
Looks like someone else got to this 3 minutes before I did (via https://github.com/react-bootstrap/react-overlays/commit/933b159bb5e3e7fe99636b49c488ec249f7bd8dd)

Thanks!

jquense:
thanks for being so speedy anyway!

evantahler:
No problem!  Thanks for the awesome software!

PR #826, opened 2020-06-09
chore: Don't deploy docs from master

taion:
cc @bpas247 @jquense 

on second thought, i don't think we should be deploying docs from master. if we do, then we have a risk of showing unreleased stuff on the docs page, which has in the past been a cause of confusion for users


mxschmitt:
> cc @bpas247 @jquense 
> 
> on second thought, i don't think we should be deploying docs from master. if we do, then we have a risk of showing unreleased stuff on the docs page, which has in the past been a cause of confusion for users

We can adjust the on-condition so that it will only be triggered on new tags or releases.

mxschmitt:
@taion adjusted so that it only deploys on releases.
(releases = GitHub releases. Haven't used tags, since tags can be re-pushed and that would cause CI to rerun)

taion:
@mxschmitt merging with your change, then

PR #837, opened 2020-07-10
feat: improve popper integration

jquense:
Should be fine to adapt in RB without a breaking change there

BREAKING CHANGE: popperConfig longer accept object forms of modifiers, pass an array instead
BREAKING CHANGE: overlay and dropdown menu injected values are different
BREAKING CHANGE: overlay no longer triggers an update when placement change due to auto or flip placements

### DIFF
--- a/src/DropdownMenu.tsx
+++ b/src/DropdownMenu.tsx
@@ -116,12 +116,12 @@ export function useDropdownMenu(options: UseDropdownMenuOptions = {}) {
       ...childArgs,
       props: {
         ...menuProps,
-        ...(attributes.popper || {}),
+        ...attributes.popper,
         style: styles.popper as any,
       },
       arrowProps: {
         ref: attachArrowRef,
-        ...(attributes.arrow || {}),
+        ...attributes.arrow,
         style: styles.arrow as any,
       },
     };
--- a/src/DropdownMenu.tsx
+++ b/src/DropdownMenu.tsx
@@ -2,18 +2,16 @@ import PropTypes from 'prop-types';
 import React, { useContext, useRef } from 'react';
 import useCallbackRef from '@restart/hooks/useCallbackRef';
 import DropdownContext from './DropdownContext';
-import usePopper, {
-  toModifierMap,
-  UsePopperOptions,
-  Placement,
-} from './usePopper';
+import usePopper, { UsePopperOptions, Placement, Offset } from './usePopper';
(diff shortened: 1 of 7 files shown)


src/DropdownMenu.tsx
```diff
@@ -125,11 +116,13 @@ export function useDropdownMenu(options: UseDropdownMenuOptions = {}) {
       ...childArgs,
       props: {
         ...menuProps,
-        style: popper.styles as any,
+        ...(attributes.popper || {}),
```
taion, line 119: **nitpick:** you shouldn't need `|| {}` here? at least `|| null` or whatever should be enough. object spreads work with null-y values. actually, it works with any falsy value, so i'm not sure why we need a fallback here at all?

src/usePopper.ts
```diff
@@ -21,53 +23,41 @@ export type Placement = Popper.Placement;
 export type VirtualElement = Popper.VirtualElement;
 export type State = Popper.State;
 
+export type OffsetsFunction = (details: {
+  popper: Popper.Rect;
+  reference: Popper.Rect;
+  placement: Placement;
+}) => [number | null | undefined, number | null | undefined];
+
+export type Offset =
```
taion, line 32: **nitpick:** `OffsetsFunction` v `Offset` is odd, no? it should be consistently singular or plural.

src/usePopper.ts
```diff
@@ -21,53 +23,41 @@ export type Placement = Popper.Placement;
...
+  | OffsetsFunction
+  | [number | null | undefined, number | null | undefined];
```
taion, line 34: **nitpick:** factor out shared `OffsetValue` / `OffsetsValue`?

src/Overlay.tsx
```diff
@@ -49,118 +53,104 @@ export interface OverlayProps extends TransitionCallbacks {
...
+    if (!mountOverlay) {
+      // Don't bother showing anything if we don't have to.
+      return null;
+    }
+
+    let child = props.children({
+      ...popper,
+      show: !!props.show,
+      props: {
+        ...(attributes.popper || {}),
```
taion, line 121: **nitpick:** ... &c.

```suggestion
        ...attributes.popper,
```

src/usePopper.ts
jquense: oops yeah, i just copied it from popper

Issue #836, opened 2020-07-10
Outer overlay(Popper) is getting closed on performing action on inner overlay(Popper) [RootClose: true for both the ovelays]


ramareddy12345:
We are opening inner overlay on button click inside outer overlay. both the overlays has Rootclose: true and both are appended to dom as siblings through portal.(container is body tag for both the overlays)

**Currentbehaviour**: useRootcloseWrapper hook doesn't have mechanism to check relate inner & outer overlay relation and considering it as different overlay and triggering onHide on outer overlay resulting in outer overlay getting closed.

Expected Behaviour:  usePopper should maintain stack of overlay hierarchy (After creating popper instance) and consider overlays relation in useRootcloseWrapper . Functionality should be similar to https://runfaj.github.io/react-nested-popper/#/nested

jquense:
if we could rely on Portal event bubbling for our click outside logic this could be neatly avoided...I don't think there is any good way of doing that tho? Or maybe we just need to mark the native event as bubbled and check it in a document listener? Didn't we try that before @taion?

taion:
i think that was the old, old implementation, and i recall it led to some weird edge cases

@ramareddy12345 why not just set the container for the inner overlay to be something inside the outer overlay? wouldn't that solve the issues here?

PR #843, opened 2020-07-21
feat: Make DropdownAPI consistent and fix keyboard handling

jquense:
Also removes the need for a wrapping element

### DIFF
--- a/src/Dropdown.tsx
+++ b/src/Dropdown.tsx
@@ -4,8 +4,8 @@ import React, { useCallback, useRef, useEffect, useMemo } from 'react';
 import PropTypes from 'prop-types';
 import { useUncontrolledProp } from 'uncontrollable';
 import usePrevious from '@restart/hooks/usePrevious';
-import useCallbackRef from '@restart/hooks/useCallbackRef';
 import useForceUpdate from '@restart/hooks/useForceUpdate';
+import useGlobalListener from '@restart/hooks/useGlobalListener';
 import useEventCallback from '@restart/hooks/useEventCallback';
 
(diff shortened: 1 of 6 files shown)


src/Dropdown.tsx
```diff
@@ -90,10 +90,24 @@ export interface DropdownProps {
...
-  onToggle: (nextShow: boolean, event?: React.SyntheticEvent) => void;
+  onToggle: (nextShow: boolean, event?: React.SyntheticEvent | Event) => void;
   itemSelector?: string;
   focusFirstItemOnShow?: false | true | 'keyboard';
-  children: (arg: { props: DropdownInjectedProps }) => React.ReactNode;
+  children: React.ReactNode;
+}
+
+function useRefWithUpdate() {
+  const forceUpdate = useForceUpdate();
```
taion, line 100: **question:** do these get properly enqueued/deduped, or could we end up with a double-rerender?

jquense: I'm fairly sure it does, maybe not a big deal either way idk

jquense:
I realized recently that we don't handle `escape` inside of a modal for dropdowns which leads to both firing. Bootstrap calls preventDefault and stopPropagation for those, but that's tricky here given that this now fires after react sees the event.

Should try and attach the events in react via props? I was trying to avoid extra event merging needed by users, but maybe that's ok here? Alternatively i could do the capture phase tagging etc

taion:
ohh that's unfortunate.

these don't even both go through root close wrapper, right?

i guess we could have some shared `WeakMap` or something there to ensure events are only processed once? would also address https://github.com/react-bootstrap/react-overlays/issues/836

PR #855, opened 2020-08-19
Removing aria-haspopup

romans12:


jquense:
should we conditionally add it for cases where the menu is actually a `role=menu`? We do some of that already, it may not be too much extra work

PR #865, opened 2020-09-06
Add wrapperRender method to Modal component

vileven:


jquense:
Hi, thanks for the PR. I'm not sure i understand what this adds or unlocks. You can already control the rendering of each as well as the portal `container`

vileven:
> Hi, thanks for the PR. I'm not sure i understand what this adds or unlocks. You can already control the rendering of each as well as the portal `container`

I need to have a wrapper, due to the design of my modals (I think it doesn't really matters why. But in case you're interested, it's because they have to add a scrollbar if they don't fit on the page).

Each modal should have a wrapper...

and of course, I can use container. But in that container, I should implement container prop pretty similar inside this library. And I get a very complex structure with two `ReactDOM.createPortal` and so on. So, I decided to fork library and add wrapper here.

jquense:
thoughts folks?

kyletsang:
Hmm I'm not sure about this.  I'm thinking it should be possible to handle the scrolling using the current rendering API, but maybe it's easier to visualize with an example?

PR #924, opened 2021-03-01
chore: bump deps

jquense:


karma.conf.js
```diff
@@ -25,18 +25,16 @@ module.exports = (config) => {
       resolve: {
         symlinks: false,
         extensions: ['.mjs', '.js', '.ts', '.tsx', '.json'],
+        // for Cheerio
+        fallback: { stream: require.resolve('stream-browserify') },
```
kyletsang, line 29: Looks like we also need to include the polyfill for `util`

kyletsang (approved):

PR #957, opened 2021-06-23
Keyboard focus error dropdown

golota60:
Resolves: https://github.com/react-bootstrap/react-bootstrap/issues/5883

Tested using next.js in development mode. (Fun fact: i've been able to reproduce this bug only using next js and only in dev mode - in production everything worked fine)


golota60:
woops, pushed from the wrong branch - sorry

PR #958, opened 2021-06-23
fix(Dropdown): add checking if ref exists on dropdown

golota60:
Resolves: https://github.com/react-bootstrap/react-bootstrap/issues/5883

Tested using next.js in development mode. (Fun fact: i've been able to reproduce this bug only using next js and only in dev mode - in production everything worked fine)

### DIFF
--- a/src/Dropdown.tsx
+++ b/src/Dropdown.tsx
@@ -278,9 +278,7 @@ function Dropdown({
           (e) => {
             if (
               (e.key === 'Tab' && !e.target) ||
-              !(
-                menuRef.current?.contains(e.target as HTMLElement)
-              )
+              !menuRef.current?.contains(e.target as HTMLElement)
             ) {
               onToggle(false, event);
             }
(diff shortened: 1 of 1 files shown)


src/Dropdown.tsx
```diff
@@ -278,7 +278,10 @@ function Dropdown({
           (e) => {
             if (
               (e.key === 'Tab' && !e.target) ||
-              !menuRef.current!.contains(e.target as HTMLElement)
+              !(
+                menuRef.current &&
+                menuRef.current.contains(e.target as HTMLElement)
```
jquense, line 283:
```suggestion
                menuRef.current?.contains(e.target as HTMLElement)
```
maybe? I think that should be equivalent

golota60: just checked, it works 😄 thanks

kyletsang: In this case, do we want to call `onToggle` if the menu ref isn't defined (menu isn't opened)?

jquense (approved):

PR #1013, opened 2022-09-11
fix(usePopper): fix exported Modifier type

### LINKED ISSUE #1012: react-overlays won't compile with typescript 4.7+
## Describe the bug

As per title, due to issue with the used version of `@popperjs/core` where the generic argument `Options` of type `Modifier` does not extend the required `Obj` type typescript will throw an error. Forcing the resolution to the `@popperjs/core@2.11.6` which fixes the issue causes a problem with `react-overlays` own types as they would be then incompatible in the `usePopper.d.ts` export declaration for the aforementioned `Modifier` type.

## To Reproduce

1. Install `typescript` v. `4.7+`
2. Run `tsc --noEmit`
3. See error

## Expected behavior

An error should not occur

## Environment (please complete the following information)

- react-overlays Version [e.g. 5.2.0]

iz-podpolja:
As per title, fixes: https://github.com/react-bootstrap/react-overlays/issues/1012

### DIFF
--- a/package.json
+++ b/package.json
@@ -58,7 +58,7 @@
   },
   "dependencies": {
     "@babel/runtime": "^7.13.8",
-    "@popperjs/core": "^2.8.6",
+    "@popperjs/core": "^2.11.6",
     "@restart/hooks": "^0.4.7",
     "@types/warning": "^3.0.0",
     "dom-helpers": "^5.2.0",
--- a/src/usePopper.ts
+++ b/src/usePopper.ts
@@ -16,7 +16,7 @@ const initialPopperStyles = (
 const disabledApplyStylesModifier = { name: 'applyStyles', enabled: false };
 
 // until docjs supports type exports...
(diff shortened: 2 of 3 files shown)


kyletsang:
Thanks for the PR.  Could you run prettier on your code?

kyletsang:
Thanks. Prettier looks good on the code, but we don't need those changes on the changelog

kyletsang:
There's also this error

```
Type 'undefined' does not satisfy the constraint 'Obj'.
```

iz-podpolja:
> There's also this error
> 
> ```
> Type 'undefined' does not satisfy the constraint 'Obj'.
> ```

Ok, excluded the `changelog.md` from prettier output and added a util to proxify `undefined` with `Popper.Obj`

kyletsang (approved):

kyletsang:
Thanks!

